Issue #22, opened 2023-11-22
Build on GitHub


ionutanin:
- [x] Manual workflow dispatch for releasing a new version
- [ ] Build extension for all browsers for every PR and delete all uploaded artifacts when the PR is closed

PR #32, opened 2023-12-05
chore(ci): add release workflow

raducristianpopa:
First part of #22. 

This PR contains a workflow that can be manually dispatched from the Actions tab, which will build the extension for all browsers, create a new release on GitHub and upload the assets for it (extension ZIP).


.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
...
+    steps:
+      - name: Check release version
+        id: release-version
+        env:
+          VERSION: ${{ env.VERSION }}
+        run: |
+          if [[ ! "${VERSION}" =~ ^[0-9]+\.[0-9]+\.[0-9]+.*$ ]]; then
+              echo "Invalid version tag '${VERSION}'"
+              exit 1
+          fi
```
raducristianpopa, line 32: Do we need a better version validator?

PR #305, opened 2024-06-05
fix: deduplicate token rotation requests

### LINKED ISSUE #288: [Background] Handle token rotation synchronously
When there are multiple link tags on a page, each link tag may going to try to rotate its token

---

If there are multiple payment sessions for a tab and the token expires, each payment session will try to rotate the token and the monetization flow will break because of the race condition.

To be implemented:
- move token rotation logic in `OpenPaymentsService`
- dedupe token rotation calls and cache the results for a small amount of time

raducristianpopa:
## Context

Closes #288.


raducristianpopa:

<table role="table">
<thead>
  <tr>
    <th>Name</th>
    <th>Link</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td align="center">bdf4fbe90a333ca59de34ade0906cd8edd402d06</td>
  </tr>
  <tr>
    <td align="center"><a href="https://github.com/interledger/web-monetization-extension/actions/runs/9379684757">Run #9379684757</a></td>
  </tr>
  <tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/24546627974/artifacts/1569875116">Download</a></td></tr><tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/24546627974/artifacts/1569875120">Download</a></td></tr>
</tbody>
</table>

src/background/services/openPayments.ts
```diff
@@ -76,22 +79,42 @@ interface CreateQuoteAndOutgoingPaymentGrantParams {
...
-    ;(async () => {
-      const { connected, walletAddress } = await this.storage.get([
-        'connected',
-        'walletAddress'
-      ])
-      if (connected === true && walletAddress) {
-        this.initClient(walletAddress.id)
-      }
-    })()
+    console.log(this.token)
```
dianafulga, line 103: logger or delete it

src/background/services/openPayments.ts
```diff
@@ -453,4 +479,75 @@ export class OpenPaymentsService {
...
+    const newToken = {
+      value: token.access_token.value,
+      manage: token.access_token.manage
+    }
+    await this.storage.set({
+      token: newToken
+    })
+    this.token = newToken
+
+    console.log(
```
dianafulga, line 546: same

dianafulga (approved):

src/background/services/deduplicator.ts
```diff
@@ -0,0 +1,73 @@
...
+  private readonly duration = 5000
+
+  constructor(private logger: Logger) {}
+
+  dedupe<T extends AsyncFn<any>>(
+    fn: T,
+    options: DedupeOptions = { cacheFnArgs: false }
+  ): T {
+    return ((...args: Parameters<T>): ReturnType<T> => {
+      const key = this.generateCacheKey(fn, args, options.cacheFnArgs)
```
sidvishnoi, line 24: Should we make this cache get an option to caller instead? I think there can be cases when JSON.stringify all args could cause performance issues if called often. We could maybe use: `{key: () => fn.name}` as default.

`wait` should be an option to be passed too.

src/background/services/openPayments.ts
```diff
@@ -453,4 +480,66 @@ export class OpenPaymentsService {
...
+    await this.storage.set({
+      token: {
+        value: token.access_token.value,
+        manage: token.access_token.manage
+      }
+    })
+    this.token = {
+      value: token.access_token.value,
+      manage: token.access_token.manage
+    }
```
sidvishnoi, line 543: Nit: Let's set `this.token` first and reuse in `this.storage.set` to avoid duplication. If concerned about `storage.set` throwing and `this.token` being inconsistent, then can assign to a new variable first.

src/background/services/paymentSession.ts
```diff
@@ -28,8 +27,7 @@ export class PaymentSession {
     private tabId: number,
     private frameId: number,
     private rate: string,
-    private openPaymentsService: OpenPaymentsService,
-    private storage: StorageService
+    private openPaymentsService: OpenPaymentsService
```
sidvishnoi, line 30: Nit: We should probably enable trailingComma in prettier.

src/background/services/paymentSession.ts
```diff
@@ -120,76 +118,34 @@ export class PaymentSession {
...
-                assetCode: walletAddress.assetCode
-              }
-            }
-          )
+        // Quote can be removed once the Test Wallet upgrades to alpha-10.
+        // We will be able to create an outgoing payment with an incoming payment,
+        // making the quoting unnecessary through OP.
+        //
+        // Note: Under the hood, Rafiki is still quoting.
+       if (!quote) {
```
sidvishnoi, line 133: Why do we have this `if (!quote)`, which is always true? We just defined it as such above.

src/background/services/paymentSession.ts
```diff
@@ -288,52 +243,23 @@ export class PaymentSession {
...
-            quoteId: quote.id,
-            metadata: {
-              source: 'Web Monetization'
-            }
-          }
-        )
+      outgoingPayment = await this.openPaymentsService.createOutgoingPayment({
+        walletAddress: this.sender,
+        quoteId: quote.id
+      })
```
sidvishnoi, line 262: Perfect 💯

sidvishnoi (dismissed):
Some nits. Looks good.

src/background/services/paymentSession.ts
sidvishnoi: But we've:
```js
while (this.active) {
      let quote: Quote | undefined
      try {
```

If `quote` is not created, `this.openPaymentsService.createQuote` will throw, right? That means, `quote` isn't assigned and we get into catch block. `quote` is not used outside of that try block from what I see.

src/background/services/deduplicator.ts
raducristianpopa: `fn.name` is already used as the default, if we do not want to cache with the args.

I like the approach of having the `wait` as an option when calling dedupe, instead of having it as a private field in the class.

sidvishnoi (approved):

PR #307, opened 2024-06-06
feat: Add intent to background & split amount

### LINKED ISSUE #285: [Background] Split sent amount
Context:
In alpha when there are multiple monetization links in a page, we sent the rate of pay that the user sets up in the interface, to each link.

Desired behaviour:
The rate of pay that is setup by the user in the interface should be divided by number of monetized links in the page, sent that amount to each link.

Implementation details:
If there are iframes in the page this is going to be the behaviour:
- the main frame will send the links found in the page => the monetization will start on each link from the main frame with rateOfPay divided by number of links in the main frame
- the first iframe will send the links found in the iframe => the monetization will start on each link from the iframe with rateOfPay divided by the number of links in main frame + first iframe, the existing monetization links from the main frame will have the amount updated to this amount
- the second iframe will send the links found in the iframe => the monetization will start on each link from the iframe with rateOfPay divided by the number of links in main frame + first iframe + second iframe, the existing monetization links from the main frame + first iframe will have the amount updated to this amount
- and so on, for each iframe found in the page

There is a possibility for a few seconds, depending on how many links are in the page) to have different amounts send to links, until we parse all links.

### LINKED ISSUE #301: [Background] Delete sessions if intent set to delete from content script
When a link tag is removed or has the href changed, the content sends the STOP_MONETIZATION event alongside an intent (delete) so that the session can be stopped and removed from memory.
Because we keep in memory all the wallet addresses details for all pages, as optimisation to have an up-to-date image of current links in a page, then we must delete from memory the details for outdated links.

dianafulga:
Closes #301 
Closes #285


raducristianpopa:

<table role="table">
<thead>
  <tr>
    <th>Name</th>
    <th>Link</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td align="center">516185c5aa906b46021e1ca20137298699c62b0c</td>
  </tr>
  <tr>
    <td align="center"><a href="https://github.com/interledger/web-monetization-extension/actions/runs/9396833026">Run #9396833026</a></td>
  </tr>
  <tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/24593961183/artifacts/1574087424">Download</a></td></tr><tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/24593961183/artifacts/1574087429">Download</a></td></tr>
</tbody>
</table>

src/background/services/monetization.ts
```diff
@@ -55,9 +54,17 @@ export class MonetizationService {
       this.sessions[tabId] = new Map()
     }
 
+    const sessionsCount = this.sessions[tabId].size + payload.length
```
sidvishnoi, line 57: Nit: How about we assign `this.sessions[tabId]` to a variable here first.

sidvishnoi, line 57: Question: shouldn't we exclude `remove` intents from `payload.length` here?
Also, do we emit stopPaymentSession before startPaymentSession so existing sessions get removed first?

src/background/services/monetization.ts
```diff
@@ -49,16 +50,17 @@ export class MonetizationService {
...
     const { tabId, frameId } = getSender(sender)
+    const sessions = this.sessions[tabId]
 
-    if (this.sessions[tabId] == null) {
+    if (sessions == null) {
       this.sessions[tabId] = new Map()
     }
 
-    const sessionsCount = this.sessions[tabId].size + payload.length
+    const sessionsCount = sessions.size + payload.length
```
sidvishnoi, line 59: Bug: If `sessions` is not set initially, we're not setting it to `new Map()` in the if block. Consider assigning `const sessions` after the if block.

src/content/services/monetizationTagManager.ts
```diff
@@ -216,8 +216,8 @@ export class MonetizationTagManager extends EventEmitter {
       }
     }
 
-    this.sendStartMonetization(startMonetizationTags)
     this.sendStopMonetization(stopMonetizationTags)
+    this.sendStartMonetization(startMonetizationTags)
```
sidvishnoi, line 220: Question: I think we should `await` the stop and then run the start. I'm worried about race conditions.

sidvishnoi (changes requested):
A little bug to squash there..

sidvishnoi (approved):

Issue #23, opened 2023-11-22
Move to MV3 as MV2 is going to be deprecated


ionutanin:
CORS on MV3
Use MonetizationEvent class rather than using custom events
Add script injection into content script on MV3 with monetization event (if necessary add script CSP sha)

- The component that searches for WM tags and event routing to WM tags should not run in the main JS context

**References on Chrome extension MV2 sunset timeline**: 
- https://developer.chrome.com/docs/extensions/develop/migrate/mv2-deprecation-timeline
- https://developer.chrome.com/blog/resuming-the-transition-to-mv3

tselit:
Prioritising compatibility for Chrome first.
Agreed to work on compatibility for other engines after work-week.

tselit:
Diana will investigate the approach that we'll need to ensure that the extension's polyfill capability is secure once we are MV3 compatible

There seems to be an issue with Firefox browser compatibility that will be an impediment for the MV3 update for the Firefox build.

sidvishnoi:
Also relevant bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1573659
Firefox MV3 doesn't support service worker as of now. We'll need separate implementation wrappers as background script/service worker.

tselit:
> Another difference we need to handle: By default, MV3 disables our all hosts permissions. That's something we need to request user, otherwise much of the extension won't work - can only open the popup, no payments or anything. Unfortunately, can't ask that permission on install either - needs user interaction. So, I'll see if we do something in UI (popup), or do it before we even show allow opening popup (less UI work that way).

Interesting, thanks Sid. I'd love to see a demo of the alternatives, actually.

sidvishnoi:
Also, not a blocker, but Firefox doesn't support content script execution world yet as mentioned in https://github.com/interledger/web-monetization-extension/issues/9.
https://bugzilla.mozilla.org/show_bug.cgi?id=1736575
So, we'll have to stick with the inject `script.innerHTML` hack for now.

sidvishnoi:
Ok, found the best bug yet. Thanks to webpack (or our configuration). It adds a `document` check to `background.js`, but Chrome MV3 (with Service Worker) doesn't support document, so I kept getting `document is not defined` errors from somewhere. 

Stack trace pointed only to our logger, so had to grep for document in bundles. Will check how do we avoid webpack adding that.

Rest of the extension seems to be working fine with Firefox/Chrome MV3 🎉

[...]

PR #313, opened 2024-06-07
feat!: support manifest v3

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/23

Changes proposed in this pull request
- Use updated `manifestv3.json`, and modify it as needed for Firefox/Chrome support.
- As MV3 doesn't enable `hosts` permissions on install (automatically or manually), show an error message that permissions are required, along with a button to request permission. No other part of UI is shown until we've the permissions.
- Make Webpack generate self-contained bundles. Otherwise, it uses dynamic script loading with `document.createElement('script')`, and `document` isn't defined in MV3 background service workers, which causes a fatal crash.
- Clear all install-time warnings in Firefox as well as Chrome.
- In MV3, production extension will only work on https:// websites. ([See comment](https://github.com/interledger/web-monetization-extension/pull/313#discussion_r1632601314))

---

Questions:
- [x] Should I get rid of manifest v2 stuff? I tested it works fine in Firefox 110 even.
     - ~Decided to retain functionality, with a `--mv2` flag during build.~
     - Decided later in meeting to remove v2 stuff.
- [x]  Should I show the permission error/prompt on all screens?
     - Decided in meeting we'll show only the error if we don't have the permissions.


webpack/plugins.ts
```diff
@@ -14,6 +14,9 @@ export const getMainPlugins = (outputDir: string, target: Target): any[] => [
     verbose: true
   }),
   new ProgressPlugin(),
+  new optimize.LimitChunkCountPlugin({
+    maxChunks: 1
+  }),
```
sidvishnoi, line 19: Chrome MV3 broke without this. As webpack tried dynamic chunk loading with `document.createElement('script')` mechanism. So, told webpack bundle everything into single file.
I haven't used webpack since long, but I think there must be some mechanism to dynamic load? Single chunk works fine too anyway.

webpack/plugins.ts
```diff
@@ -32,12 +35,21 @@ export const getMainPlugins = (outputDir: string, target: Target): any[] => [
...
+          return path.resolve(ROOT_DIR, `${outputDir}/${target}/manifest.json`)
+        },
+        transform(content: Buffer) {
+          const json = JSON.parse(content.toString())
+          delete json['$schema']
           if (target === 'firefox') {
-            // TODO: Update manifest for Firefox (V3)
+            json.background = {
+              scripts: [json.background.service_worker]
+            }
```
sidvishnoi, line 47: This was mostly it to support it in Firefox 🎉

src/manifestv3.json
```diff
@@ -11,19 +12,20 @@
...
       "matches": ["http://*/*", "https://*/*"],
-      "js": ["content/content.js"]
+      "js": ["content/content.js"],
+      "all_frames": true
     },
     {
       "run_at": "document_start",
       "matches": ["http://*/*", "https://*/*"],
-      "js": ["contentStatic/contentStatic.js"]
+      "js": ["contentStatic/contentStatic.js"],
```
sidvishnoi, line 21: We get CSP failure in Chrome at the moment. 
And Chrome doesn't allow lowering `content_security_policy.extension_pages` in any way (no nonce or SHA allowed).

Will need an approach with injectScript or web_accessible_resources I think.

src/popup/components/ConnectWalletForm.tsx
```diff
@@ -76,6 +97,31 @@ export const ConnectWalletForm = ({ publicKey }: ConnectWalletFormProps) => {
       })}
       className="space-y-4"
     >
+      {errors.root && (
```
sidvishnoi, line 100: Looks like this
image

src/_locales/en/messages.json
```diff
@@ -10,5 +10,9 @@
   "btnTooltip": {
     "message": "Web Monetization",
     "description": "..."
+  },
+  "hostsPermissionsNeeded": {
+    "message": "Permission to access your data for all websites is required to use Web Monetization.",
```
sidvishnoi, line 15: Language copied from Firefox permissions page.
> Access your data for all websites

src/manifestv3.json
raducristianpopa, line 1: For the production build, should we restrict the extension to only run in a secure context?

Based on the WM Spec, the interfaces should only be present in a secure context: https://webmonetization.org/specification/#monetizationevent-interface.

src/manifestv3.json
```diff
@@ -35,9 +38,19 @@
...
+        "background/*",
+        "specs/*"
       ],
       "matches": ["<all_urls>"]
     }
-  ]
+  ],
+  "content_security_policy": {
+    "extension_pages": "script-src 'self'; object-src 'self'"
+  },
```
raducristianpopa, line 49: There is no need to generate a nonce?

src/manifestv3.json
```diff
@@ -11,19 +12,21 @@
...
       "matches": ["http://*/*", "https://*/*"],
-      "js": ["content/content.js"]
+      "js": ["content/content.js"],
+      "all_frames": true
     },
     {
       "run_at": "document_start",
       "matches": ["http://*/*", "https://*/*"],
-      "js": ["contentStatic/contentStatic.js"]
+      "js": ["contentStatic/contentStatic.js"],
```
raducristianpopa, line 21: Not really important at the moment, but I find the `contentStatic` naming a little but confusing. Probably can be renamed to `polyfill`/`wm-polyfill`.

src/popup/components/ConnectWalletForm.tsx
raducristianpopa: This is only going to happen if the user installs the extension from Chrome Web Store or Mozilla Add-ons and declines the popup for permissions?

raducristianpopa:
>     1. Should I get rid of manifest v2 stuff? I tested it works fine in Firefox 110 even.

I would keep them for now. We can probably default to using MV3 (renaming `manifestv3.json` to `manifest.json`) and keep the old manifest for the moment.

>     2. Should I show the permission error/prompt on all screens?

i don't think so, since the button will be disabled and we check the permission when the user submits the form.

src/manifestv3.json
sidvishnoi: It's in fact, disallowed from what I tried. Leads to invalid CSP errors in Chrome (so does using a SHA).

src/manifestv3.json
sidvishnoi: Yeah, can do as follow-up. Will prefer to keep PR as small as possible.

sidvishnoi:
> > Should I show the permission error/prompt on all screens?
>
> i don't think so, since the button will be disabled and we check the permission when the user submits the form.

So, content script also won't be injected if current URL isn't doesn't have the permission. We ask for the permission only on connect screen, but if some has connected before, and they update the extension, they won't see that they haven't given the permissions.
Also valid in case the permission is removed later (for some reason - like mistake or browsers being more strict in future with privacy considerations, like removing permission after N days).

src/popup/components/ConnectWalletForm.tsx
sidvishnoi: It's going to happen even with load-unpacked extension. And we can't ask for the hosts permission at install time. So, this will be visible as whenever we don't have the permissions. We should perhaps add another icon or reuse some existing when we don't have the permissions to show action is _disabled_, and prompt user to give permission.

dianafulga: having a scroll bar in the popup I would say is a pretty bad UX, maybe just display this message, and just after that let the user config things? To be discussed with the team/designer

sidvishnoi: Show only the message when missing permissions?

dianafulga: The UI to be addressed in a dedicated ticket

dianafulga (approved):

src/popup/components/layout/MainLayout.tsx
```diff
@@ -1,20 +1,45 @@
...
+    void checkPermissions()
+    browser.permissions.onAdded.addListener(checkPermissions)
+    browser.permissions.onRemoved.addListener(checkPermissions)
+
+    return () => {
+      browser.permissions.onAdded.removeListener(checkPermissions)
+      browser.permissions.onRemoved.removeListener(checkPermissions)
+    }
+  }, [location])
+
```
sidvishnoi, line 37: I'm not sure we should do these checks in popup or pass these through background (popup state). @dianafulga?

src/popup/components/layout/MainLayout.tsx
```diff
@@ -1,20 +1,44 @@
...
+import { MissingHostPermission } from '@/popup/components/MissingHostPermission'
 
 const Divider = () => {
   return <div className="w-100 h-1 bg-divider-gradient" />
 }
 
 export const MainLayout = () => {
+  const [hasHostPermission, setHasHostPermission] = React.useState(true)
+
+  React.useEffect(() => {
```
sidvishnoi, line 17: I'm not sure we should do these checks in popup or pass these through background (popup state). @dianafulga?

dianafulga: I would pass it through the background, because:
1. we might want to add in the feature logic for a certain icon to be displayed
2. I would keep the browser usage from polyfill limited to background, to easily change it in the feature. This is why we actually inject it in the container in the first place

sidvishnoi: > keep the browser usage from polyfill limited to background

Though, we'd eventually need it in popup, for example `browser.i18n.getMessage`, as well as for requesting permission (on user click). Maybe I can remove the polyfill there and add a lighter wrapper for these two at least. We don't need most of the polyfill there.

dianafulga (approved):

Issue #292, opened 2024-05-31
False-positive 1-time payment success msg


tselit:
### Steps to reproduce

**Wallet setup**
1. Successfully created a USD sender test wallet account in rafiki.money.
2. Successfully connected the wallet address to the extension.

**Visited websites**
1. Visited radu.sh and the WM icon was green
2. Successfully made a one-time payment to [radu.sh](radu.sh) (so far, so good :smiley: )

3. Visited [google.com](google.com) (a non-WMized website) and the WM icon remained red 
5. I opened the WM ext, entered an amount, and attempted to make a 1-time payment to google.com.
6. The ext said "This website is not monetized."   (still going well :smiley: )

7. While I still had google.com open, with the WM ext open, and a value on the page for sending a 1-time payment
8. I used a keyboard shortcut to open the browser dev tools (**note:** the Web browser window remained open on the side, with the extension pop-up also **still** open) and I clicked around in the dev tools window
9. I then moved the mouse back to the Web browser, and clicked "Send Money" on the extension
10. The extension showed the message "payment successful"   (this can't be 👎🏾 )

### Expected result

No payments can happen, and no payment success messages can be valid when attempting to pay a non-monetized website with the extension.

### Actual result

The extension indicated that a one-time payment to google.com had succeeded.

### Screenshots or videos

**Link to video**: [on loom](https://www.loom.com/share/07ca89dcfd094e89af40e3408be0ab13?sid=dfa6ab83-ccbb-47d6-be0b-c229a0c9183b)

Was able to replicate behaviour on dev.to:
![image](https://github.com/interledger/web-monetization-extension/assets/90185897/6b250f5e-ab30-4819-b21c-afb1f2c77851)

### Additional context
We also noticed that there are instances where a website has multiple valid wallet addresses, and a one-time payment only succeeds in that some of the wallet addresses receive payment, and others do not (for example, if the sending wallet exceeds the authorized amount after successfully making 1 of 2 payments).

_No response_

### Operating system

Windows

### Operating system version

10

### Browsers

Chrome

### Browser version

125.0.6422.78

### Extension version

alpha 1 (the ext shows v1.0.1)

tselit:
**To do**: @tselit to log a separate ticket (using the same video attached to this ticket, re. concatenated text on the "Send now" button.

So this is ready for review.

raducristianpopa:
Closed by #333.

PR #333, opened 2024-06-13
fix: use window ID when querying active tab

raducristianpopa:
## Context

Part of #292.

There was another edge case, but the video is not available anymore. It can be fixed in another PR or we keep this as draft until then.

Aparently this is a known bug: https://issues.chromium.org/issues/41160154.

In the video that was provided in the initial issue, when the extension popup is open and the focus is on the DevTools. moving focus to the extension's popup, we will get an empty array when we query the active tab.

The false positive is happening because of this return statement:
https://github.com/interledger/web-monetization-extension/blob/bf1baff5802a4472045c1bcca094873b3f4b919a/src/background/services/monetization.ts#L201-L203

Any suggestions for a good error message?


raducristianpopa:

<table role="table">
<thead>
  <tr>
    <th>Name</th>
    <th>Link</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td align="center">0408491d5352be5a82e5f443518809abc9714bc4</td>
  </tr>
  <tr>
    <td align="center"><a href="https://github.com/interledger/web-monetization-extension/actions/runs/9502392977">Run #9502392977</a></td>
  </tr>
  <tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/24872749376/artifacts/1598673654">Download</a></td></tr><tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/24872749376/artifacts/1598673657">Download</a></td></tr>
</tbody>
</table>

src/background/services/storage.ts
```diff
@@ -71,7 +71,7 @@ export class StorageService {
     if (tab && tab.url) {
       try {
         const tabUrl = new URL(tab.url)
-        if (tabUrl.protocol === 'https:') {
+        if (tabUrl.protocol === 'http:') {
```
sidvishnoi, line 74: Why not both http and https?

raducristianpopa: Just for testing purposes at the moment, since I was testing this on a local website.

Thoughts on moving the host permissions here

https://github.com/interledger/web-monetization-extension/blob/992b3460c0dce051b182f0d1a169535d2283a25d/src/shared/defines.ts#L1-L5

and define them in the webpack config based on the environment ala:

https://github.com/interledger/web-monetization-extension/blob/992b3460c0dce051b182f0d1a169535d2283a25d/webpack/dev.ts#L42-L44

raducristianpopa: I thought I can use that for the protocol check here as well, but they are in a different format.

sidvishnoi: Yeah this place could be better. It's in shared/constants right now, and uses process.env to switch values.

raducristianpopa: Should we add two new defines?
- PERMISSION_HOSTS
- ALLOWED_PROTOCOLS

sidvishnoi: Yes, better than trying to parse/convert each for different uses.

raducristianpopa: Let me do this in a separate PR.

sidvishnoi: I'd update plugins script to use same constants to modify manifest.json if possible.

dianafulga (dismissed):

sidvishnoi (dismissed):

raducristianpopa:
Any suggestions for a good error message in case this is going to happen again? The false positive was happening because we did not throw an error when the tab ID was not defined, here:

https://github.com/interledger/web-monetization-extension/blob/740821ab1633fa14ac6125b433afa13ff81e3c82/src/background/services/monetization.ts#L203

sidvishnoi:
Re https://github.com/interledger/web-monetization-extension/pull/333#issuecomment-2178059932

Does this happen only when devtools are open? When the tab.id may be undefined?

raducristianpopa:
> Re [#333 (comment)](https://github.com/interledger/web-monetization-extension/pull/333#issuecomment-2178059932)
> 
> Does this happen only when devtools are open? When the tab.id may be undefined?

At one point I was able to reproduce it when switching between different window applications (another browser window or another app).

sidvishnoi:
@raducristianpopa I could not reproduce this. Maybe we can add a simple error like "Could not find active tab" so we know what caused it in future.

sidvishnoi:
Unrelated/related: I get this randomly sometimes
image

`sessions` undefined for some reason (just for additional safety, that check should be on `sessions?.size`, so we don't accidentally get divide by zero error.
https://github.com/interledger/web-monetization-extension/blob/740821ab1633fa14ac6125b433afa13ff81e3c82/src/background/services/monetization.ts#L207-L212

src/background/services/monetization.ts
```diff
@@ -200,11 +200,13 @@ export class MonetizationService {
...
     const tab = await getCurrentActiveTab(this.browser)
-    if (!tab || !tab.id) return
+    if (!tab || !tab.id) {
+      throw new Error('Could not find active tab.')
+    }
 
     const sessions = this.sessions[tab.id]
 
-    if (!sessions) {
+    if (sessions?.size) {
```
sidvishnoi, line 209:
```suggestion
    if (!sessions?.size) {
```

sidvishnoi (approved):

PR #341, opened 2024-06-14
chore: use webpack defines for hosts and protocols; remove profiler

### LINKED ISSUE #339: Define host permissions and allowed protocols in webpack config
Just for testing purposes at the moment, since I was testing this on a local website.

Thoughts on moving the host permissions here

https://github.com/interledger/web-monetization-extension/blob/992b3460c0dce051b182f0d1a169535d2283a25d/src/shared/defines.ts#L1-L5

and define them in the webpack config based on the environment ala:

https://github.com/interledger/web-monetization-extension/blob/992b3460c0dce051b182f0d1a169535d2283a25d/webpack/dev.ts#L42-L44

_Originally posted by @raducristianpopa in https://github.com/interledger/web-monetization-extension/pull/333#discussion_r1639612349_

raducristianpopa:
## Context

Closes #339.

Additional changes:
- removed profiler script - obsolete


webpack/dev.ts
```diff
@@ -40,8 +44,17 @@ export const getDevConfig = (target: Target): Configuration => {
...
+        CONFIG_PERMISSION_HOSTS: JSON.stringify({
+          origins: ['http://*/*', 'https://*/*']
+        }),
+        CONFIG_ALLOWED_PROTOCOLS: JSON.stringify(['http:', 'https:'])
+      }),
+      profiler &&
+        new BundleAnalyzerPlugin({
+          analyzerMode: 'server',
+          analyzerPort: target === 'firefox' ? 8888 : 8889
+        })
```
sidvishnoi, line 57: We were earlier using profiler in production only. Should add in both profiles I think. Then we can add it to `webpack/plugins.ts` and pass the flag there.

scripts/dev.ts
```diff
@@ -4,6 +4,7 @@ import { getDevConfig } from '../webpack/dev'
 import { callbackFn } from '../webpack/config'
 
 const TARGET = process.argv[2] || 'chrome'
+const PROFILER = process.argv.slice(2).includes('--profiler') ? true : false
```
sidvishnoi, line 7: Nit
```suggestion
const PROFILER = process.argv.slice(2).includes('--profiler')
```

sidvishnoi (approved):

Issue #9, opened 2023-08-28
CustomEvent properties inside `detail` won't match spec


sublimator:
### Problem Description

When using the `CustomEvent` object in the Web Monetization Polyfill, all properties are bundled inside `event.detail`. This behavior does not align with the Web Monetization specification, which expects these properties to be directly accessible on the event object.

### Suggested Solution: Using Event Subclass

To conform to the spec, an `Event` subclass can be created. However, this subclass must be created in the [main page JS context](https://developer.chrome.com/docs/extensions/reference/scripting/#type-ExecutionWorld).

---

### MV2 Implementation Details: Limitations and Workarounds

In MV2, directly injecting a `<script>` tag with `src=...` into the main document will not work reliably. The script will load asynchronously and may not execute before it's actually needed or used.

The workaround is to inject the script content directly, ensuring synchronous loading and execution. Here's a simplified version of the tested code for this:

```javascript
function injectCode(code: string) {
  inject(script => (script.innerHTML = code));
}

function inject(configure: (script: HTMLScriptElement) => void) {
  const script = document.createElement('script');
  configure(script);
  document.documentElement.appendChild(script);
  document.documentElement.removeChild(script); // Clean up
}
```

#### Handling CSP in MV2

To handle [CSP issues](https://github.com/interledger/web-monetization-projects/issues/1248), you can include a `content_security_policy` in the `manifest.json`:

```json
"content_security_policy": "script-src 'self' 'sha256-POLYFILL-HASH='; object-src 'self'"
```

In your build script:

```javascript
const data = Buffer.from(polyfill.content, 'utf-8');
const digest = createHash('sha256').update(data).digest();
const polyfillHash = `sha256-${digest.toString('base64')}`;

v2.content_security_policy = v2.content_security_policy.replace(
  'sha256-POLYFILL-HASH=',
  polyfillHash
);
```

---

### MV3 Implementation Details

In MV3, the `scripting` API from Chrome can be used to register a content script that runs in the main page JS context. Below is a sample TypeScript code snippet:

```typescript
chrome.scripting.registerContentScripts([
  {
    id: 'wm-polyfill.js',  // Unique identifier for the content script
    matches: ['https://*/*', 'http://*/*'],  // URL patterns where the script will run
    js: ['wm-polyfill.js'],  // JavaScript file to be injected
    allFrames: true,  // Run the script in all frames
    world: 'MAIN',  // Execute in the main world
    runAt: 'document_start'  // Timing for script injection
[...]

PR #346, opened 2024-06-17
fix(content): dispatch MonetizationEvent instead of CustomEvent

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/9

1. Dispatch `MonetizationEvent` instead of `CustomEvent` with detail
2. Keep supporting `event.detail` access for backward compat, but add a warning on accessing for first time.
3. Rename internal event (for communication between content scripts) from `monetization-v2` to `__wm-ext-monetization`, along with `onmonetization-attribute-changed`.
4. Rename contentScript to polyfill
5. Turn contentScript code from  a string to regular TS file, so we can typecheck it like rest of code (~presently typechecking disabled on that file as out of scope of this PR~ added typechecking too)
6. Now, MonetizationEvent interface is available on global.

A sample code:
```html
<!doctype html>
<html lang="en">
  <head onmonetization="console.warn('head', event)">
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>Web Monetization test</title>
    <link
      rel="monetization"
      href="https://ilp.rafiki.money/sidvishnoi-usd-2"
      onmonetization="console.log('inline', event)"
    />
  </head>
  <body >
    <h1>Web Monetization test</h1>
    <script>
      const link = document.querySelector('link[rel="monetization"]')
      // link.addEventListener('load', console.log)
      link.addEventListener('monetization', (event) => {
        console.log('1: event', event)
        console.log('2: event', event instanceof window.MonetizationEvent)
        console.log('3: event.detail', event.detail)
      })
      window.addEventListener('monetization', (ev) =>
        console.log('window:', ev)
      )
    </script>
  </body>
</html>
```


raducristianpopa:

<table role="table">
<thead>
  <tr>
    <th>Name</th>
    <th>Link</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td align="center">3b1224a7388d5ae56dcaf3d990819e8a2f2734ba</td>
  </tr>
  <tr>
    <td align="center"><a href="https://github.com/interledger/web-monetization-extension/actions/runs/9546766229">Run #9546766229</a></td>
  </tr>
  <tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/24983300393/artifacts/1608182393">Download</a></td></tr><tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/24983300393/artifacts/1608182396">Download</a></td></tr>
</tbody>
</table>

sidvishnoi:
Question:
https://webmonetization.org/docs/references/monetizationevent/ says we emit `amountSent` with value and currency. There's no assetScale there.
And how's `amountSent` different from `receiveAmount`?

raducristianpopa:
> Question: https://webmonetization.org/docs/references/monetizationevent/ says we emit `amountSent` with value and currency. There's no assetScale there. And how's `amountSent` different from `receiveAmount`?

The current implementation is not following the spec. Based on the spec (https://webmonetization.org/specification/#amountsent-attribute), the currency should only be an [ISO4217](https://www.iso.org/iso-4217-currency-codes.html) string. I forgot about this part when sending the details about a monetization event from the background to the content script. We should convert the amount value to a user readable format:

```ts
// We should be careful about floating point precision as well
const amountSent = receiveAmount.value * 10 ** -receiveAmount.assetScale
```

The monetization event should only contain the following properties:
- `amountSent` - based on the [PaymentCurrencyAcomunt](https://www.w3.org/TR/payment-request/#dom-paymentcurrencyamount)
- `paymentPointer`
- `incomingPayment`

`receiveAmount` should not be present.

sidvishnoi:
> the currency should only be an [ISO4217](https://www.iso.org/iso-4217-currency-codes.html) string

I think then the concern was crypto currencies not having those codes? Maybe something to figure out later?

raducristianpopa:
> > the currency should only be an [ISO4217](https://www.iso.org/iso-4217-currency-codes.html) string
> 
> I think then the concern was crypto currencies not having those codes? Maybe something to figure out later?

This shouldn't really be an issue at the moment. Most ASE that enable WM for their user will probably run Rafiki, which has a constraint on the asset code as well - ISO4217: https://rafiki.dev/concepts/asset/.

src/content/services/contentScript.ts
```diff
@@ -58,4 +59,14 @@ export class ContentScript {
       }
     )
   }
+
+  // TODO: When Firefox supports `world: MAIN`, inject this directly via
```
raducristianpopa, line 63: It looks like we might get this in the next release - Firefox 128? :eyes: 
https://bugzilla.mozilla.org/show_bug.cgi?id=1736575

What's going to be the issue if we add it with `world: 'MAIN'` right now?

src/content/polyfill.ts
```diff
@@ -0,0 +1,116 @@
...
+      if (!eventDetailDeprecationEmitted) {
+        const msg = `MonetizationEvent.detail is deprecated. Access attributes directly instead.`
+        // eslint-disable-next-line no-console
+        console.warn(msg)
+        eventDetailDeprecationEmitted = true
+      }
+      const { receiveAmount, incomingPayment, paymentPointer } = this
+      return { receiveAmount, incomingPayment, paymentPointer }
+    }
+  }
```
raducristianpopa: Are there any tools that are using `event.details` to retrieve the information?

src/content/services/contentScript.ts
sidvishnoi: Yes, I'm following that issue.

If we add `world: MAIN` now: install time warnings and `window.Monetization` not available.

sidvishnoi: (But we plan to support some older versions I think, so, might have to wait a year). I'll add the bugzilla link in a comment.

src/content/polyfill.ts
sidvishnoi: I'm not aware, but to be on safer side.

src/content/services/monetizationTagManager.ts
```diff
@@ -69,12 +70,18 @@ export class MonetizationTagManager extends EventEmitter {
...
-      const customEvent = new CustomEvent('monetization', {
-        bubbles: true,
-        detail: mozClone(details, this.document)
-      })
-
-      tag.dispatchEvent(customEvent)
+      const { incomingPayment, paymentPointer, receiveAmount } = details
+      const amountSent: PaymentCurrencyAmount = {
+        currency: receiveAmount.assetCode,
+        value: transformBalance(receiveAmount.value, receiveAmount.assetScale)
```
raducristianpopa, line 76: Not really important, but I would prefer sending the correct details directly from background. Thoughts?

src/shared/messages.ts
```diff
@@ -131,7 +131,11 @@ export enum BackgroundToContentAction {
 
 export interface MonetizationEventPayload {
   requestId: string
-  details: any
+  details: {
```
raducristianpopa, line 134: If transforming the data in background
```suggestion
  detail: {
```

src/content/services/monetizationTagManager.ts
sidvishnoi: Oh yes, I didn't realise I was in content/

src/content/polyfill.ts
```diff
@@ -0,0 +1,123 @@
...
+      monetizationTag.dispatchEvent(new MonetizationEvent(event.detail))
+    },
+    { capture: true }
+  )
+
+  window.addEventListener(
+    'onmonetization-attr-changed',
+    (event) => {
+      // eslint-disable-next-line no-console
+      console.log('onmonetization-attr-changed', event.detail)
```
raducristianpopa, line 110: I feel like we should not log this. If we want to keep it, let's make it a debug log.

sidvishnoi: Why is this event even fired though? I just left it as is for now.

sidvishnoi: I'm yet to check how this event supposed to work. Will do within this PR then.

raducristianpopa: It is used to register the `onmonetization` event handler for an element. Every time the handler is changed, we need to re-register the function.

```html
<link rel="monetization" ... onmonetization="console.log(event)" />
```

- When the page is loaded, we register the `onmonetization` event handler;
- If the website is changing the handler, we need to override the previous one;

sidvishnoi: Removed log.
Fixed bug: make sure event is fired and caught by polyfill (timing issue)
Refactor: renamed the internal event to start with `__wm_ext`.

src/content/polyfill.ts
```diff
@@ -0,0 +1,122 @@
...
+    },
+    { capture: true }
+  )
+
+  window.addEventListener(
+    '__wm_ext_onmonetization_attr_change',
+    (event: CustomEvent<{ attribute?: string }>) => {
+      const link = event.target
+      if (!(link instanceof HTMLLinkElement)) return
+      if (!link.isConnected) return
```
raducristianpopa, line 112: The `onmonetization` event handler can be present on any element:

> The onmonetization event handler is an[ event handler content attribute](https://html.spec.whatwg.org/multipage/webappapis.html#event-handler-content-attributes) that can be applied to any [element](https://dom.spec.whatwg.org/#concept-element). The user agent uses it to notify that some link has been [monetized](https://webmonetization.org/specification/#dfn-monetized).

sidvishnoi: Oh. Fixing..

sidvishnoi: btw, monetization event is on Line 95 above.

raducristianpopa (approved):

Issue #286, opened 2024-05-23
[Background] Overpaying


dianafulga:
Ordering: this ticket should ideally follow #145 

If the receiver’s wallet cannot receive AMOUNT, we should send the minimum amount that can be sent. If the receiver’s asset scale is 2, we will not be able to send SPLITTED_AMOUNT, and we need to send the smallest unit 1 + 1 (we can not send 1 - cause: @interledger/pay library, used in Rafiki). We call this action OVERPAYING. We also need to adjust the payment interval. By default, the happy path, we send a payment every second. If overpaying, we need to adjust the interval as well: We need to send a payment every 30 seconds (dynamic - it will not always be 30 seconds, we need to calculate the interval based on the smallest amount that can be sent) to facilitate RATE_OF_PAY. 

- set the amount as minimum in the constructor or when recalculating the rate.
- if overpaying, calculate the interval as well:  0.02 * 3600/0.60 = 120s (2 minutes)

When a payment session is overpaying we need to store the last payment timestamp. Reasoning: if a user keeps refreshing a page and we have a payment session that is overpaying, we will send $0.02 whenever a payment session is established for the wallet address . If we have the timestamp, we can check if the page has received a payment before and if CURRENT_TIMSTAMP is not greater than the payment timestamp we will wait for CURRENT_TIMESTAMP - PAYMENT_TIMESTAMP until sending the next payment.

Whenever a payment session is overpaying save:
- Website url (we store the sha1 the URL)
- Tab ID
- Wallet address
- Last payment timestamp
- Expires At = Last payment timestamp + MS (TBD)

Whenever we receive the START_MONETIZATION event, we check the overpaying state for the current tab + website url. If a record is found, initialize a new payment session but do not start it. We need to wait for LAST_PAYMENT_TIMESTAMP + INTERVAL (dynamic for each payment session that is overpaying) - CURRENT_TIMESTAMP

tselit:
The team revised the approach for this which impacts the estimated effort = adding another 32 days = 72 hrs (noting that Monday is a national holiday in Romania). Aiming to put into review by Thurs, 27 June.

sidvishnoi:
We can't use `Tab` as the references aren't stable across calls. Need to fall back to `tabId` approach here, and add a cleanup process. See https://github.com/interledger/web-monetization-extension/pull/420#issuecomment-2225770302

Issue #287, opened 2024-05-23
[Background] [TO REFINE] Clear outdated data regarding overpaying 


dianafulga:
Every X seconds, run a function that checks the expires at property for an overpaying session. If CURRENT_TIMESTAMP > EXPIRES_AT delete the session from state.

Related to #286

tselit:
The team will have an online tech. conversation about the tech implementation options for this.

PR #349, opened 2024-06-18
Overpaying

dianafulga:
#286

### DIFF
--- a/src/background/services/background.ts
+++ b/src/background/services/background.ts
@@ -167,6 +167,7 @@ export class Background {
       if (details.reason === 'install') {
         await this.storage.populate()
         await this.openPaymentsService.genererateKeys()
+        await this.storage.populateOverpayingSessions()
       }
       await this.checkPermissions()
     })
--- a/src/background/services/monetization.ts
+++ b/src/background/services/monetization.ts
@@ -49,7 +49,7 @@ export class MonetizationService {
       )
       return
     }
-    const { tabId, frameId } = getSender(sender)
+    const { tabId, frameId, url } = getSender(sender)
 
     if (this.sessions[tabId] == null) {
       this.sessions[tabId] = new Map()
@@ -75,7 +75,9 @@ export class MonetizationService {
         tabId,
         frameId,
         rate,
-        this.openPaymentsService
+        this.openPaymentsService,
+        this.storage,
+        url
       )
 
       sessions.set(requestId, session)
--- a/src/background/services/paymentSession.ts
+++ b/src/background/services/paymentSession.ts
@@ -9,6 +9,9 @@ import {
 import { OpenPaymentsClientError } from '@interledger/open-payments/dist/client'
 import { sendMonetizationEvent } from '../lib/messages'
 import { convert, sleep } from '@/shared/helpers'
+import { StorageService } from './storage'
+import { getHashUrl } from '../utils'
+import { OverpayingSession } from '@/shared/types'
 
 const DEFAULT_INTERVAL_MS = 1000
 const HOUR_MS = 3600 * 1000
@@ -27,7 +30,9 @@ export class PaymentSession {
     private tabId: number,
     private frameId: number,
     private rate: string,
-    private openPaymentsService: OpenPaymentsService
+    private openPaymentsService: OpenPaymentsService,
+    private storage: StorageService,
+    private url: string | undefined
   ) {
     this.adjustSessionAmount()
   }
@@ -125,6 +130,8 @@ export class PaymentSession {
 
     while (this.active) {
       try {
+        this.processOverpaying()
+
         // Quote can be removed once the Test Wallet upgrades to alpha-10.
         // We will be able to create an outgoing payment with an incoming payment,
         // making the quoting unnecessary through OP.
@@ -299,5 +306,62 @@ export class PaymentSession {
   private setAmount(amount: bigint): void {
     this.amount = amount.toString()
     this.intervalInMs = Number((amount * BigInt(HOUR_MS)) / BigInt(this.rate))
+  }
+
+  private async getOverpayingSession(): Promise<{
+    session: OverpayingSession | null
+    sessionIndex: number
+  }> {
+    const { overpayingSessions } = await this.storage.get([
+      'overpayingSessions'
+    ])
+
+    const hashUrl = await getHashUrl(this.url || '')
+
+    const sessionIndex = overpayingSessions.findIndex(
+      (session) =>
+        session.tabId === this.tabId &&
+        session.hashUrl === hashUrl &&
+        session.walletAddress === this.receiver
+    )
+    const session = sessionIndex > -1 ? overpayingSessions[sessionIndex] : null
+
+    return { sessionIndex, session }
+  }
+
+  private async processOverpaying(): Promise<void> {
+    const { session } = await this.getOverpayingSession()
+
+    if (session) {
+      // If session not expired yet, wait until it expires
+      const crtTimestamp = Date.now()
+      if (session.expireTimestamp > crtTimestamp) {
+        await sleep(session.expireTimestamp - crtTimestamp)
+      }
+    }
+  }
+
+  private async saveOverpaying(): Promise<void> {
+    const { overpayingSessions } = await this.storage.get([
+      'overpayingSessions'
+    ])
+
+    const hashUrl = await getHashUrl(this.url || '')
+    const crtTimestamp = Date.now()
+    const expireTimestamp = crtTimestamp + this.intervalInMs
+    const walletAddress = this.receiver
+
+    const { session, sessionIndex } = await this.getOverpayingSession()
+
+    const newSession = {
+      ...(session || { hashUrl, tabId: this.tabId }),
+      expireTimestamp,
+      walletAddress,
+      lastPaymentTimestamp: crtTimestamp
+    }
+
+    overpayingSessions[sessionIndex] = newSession
+
+    this.storage.set({ overpayingSessions })
   }
 }
--- a/src/background/utils.ts
+++ b/src/background/utils.ts
@@ -83,8 +83,17 @@ export const getTabId = (sender: Runtime.MessageSender): number => {
 export const getSender = (sender: Runtime.MessageSender) => {
   const tabId = getTabId(sender)
   const frameId = notNullOrUndef(sender.frameId, 'sender.frameId')
-  return { tabId, frameId }
+  return { tabId, frameId, url: sender.url }
 }
 
 export const computeRate = (rate: string, sessionsCount: number) =>
   (+rate / sessionsCount).toString()
+
+export const getHashUrl = async (url: string) => {
+  const encoder = new TextEncoder()
+  const hashUrl = await crypto.subtle.digest('SHA-1', encoder.encode(url))
+
+  return Array.from(new Uint8Array(hashUrl))
+    .map((v) => v.toString(16).padStart(2, '0'))
+    .join('')
+}
--- a/src/shared/types.ts
+++ b/src/shared/types.ts
@@ -28,6 +28,14 @@ export interface GrantDetails {
   continueUri: string
 }
 
+export interface OverpayingSession {
+  tabId: number
+  hashUrl: string
+  walletAddress: WalletAddress
+  lastPaymentTimestamp: number
+  expireTimestamp: number
+}
+
 export interface Storage {
   /** If web monetization is enabled */
   enabled: boolean
@@ -56,12 +64,19 @@ export interface Storage {
   publicKey: string
   privateKey: string
   keyId: string
+
+  overpayingSessions: OverpayingSession[]
 }
 export type StorageKey = keyof Storage
 
 export type PopupStore = Omit<
   Storage,
-  'privateKey' | 'keyId' | 'exceptionList' | 'token' | 'grant'
+  | 'privateKey'
+  | 'keyId'
+  | 'exceptionList'
+  | 'token'
+  | 'grant'
+  | 'overpayingSessions'
 > & {
   url: string | undefined
 }
--- a/src/background/services/storage.ts
+++ b/src/background/services/storage.ts
@@ -15,7 +15,10 @@ const defaultStorage = {
   rateOfPay: null,
   minRateOfPay: null,
   maxRateOfPay: null
-} satisfies Omit<Storage, 'publicKey' | 'privateKey' | 'keyId'>
+} satisfies Omit<
+  Storage,
+  'publicKey' | 'privateKey' | 'keyId' | 'overpayingSessions'
+>
 
 // TODO: Emit events when certain values are updated:
 // Eg:
@@ -103,6 +106,13 @@ export class StorageService {
     }
 
     return false
+  }
+
+  async populateOverpayingSessions(): Promise<void> {
+    const { overpayingSessions } = await this.get(['overpayingSessions'])
+    if (overpayingSessions) return
+
+    this.set({ overpayingSessions: [] })
   }
 
   async setHostPermissionStatus(status: boolean): Promise<void> {


src/background/services/background.ts
```diff
@@ -167,6 +167,7 @@ export class Background {
       if (details.reason === 'install') {
         await this.storage.populate()
         await this.openPaymentsService.genererateKeys()
+        await this.storage.populateOverpayingSessions()
```
raducristianpopa, line 170: Instead of saving them in the extension storage, shouldn't we save them in a state service in the background?

src/background/services/paymentSession.ts
```diff
@@ -300,4 +307,61 @@ export class PaymentSession {
...
+
+  private async getOverpayingSession(): Promise<{
+    session: OverpayingSession | null
+    sessionIndex: number
+  }> {
+    const { overpayingSessions } = await this.storage.get([
+      'overpayingSessions'
+    ])
+
+    const hashUrl = await getHashUrl(this.url || '')
```
raducristianpopa, line 319: Let's exclude query params from the URL, just like we are going to do for the exceptions list.

src/background/services/background.ts
sidvishnoi: Just in case, might not work out to store it in memory, as background service worker is killed after some inactivity.
But `storage.session` may work well enough?

src/background/services/paymentSession.ts
sidvishnoi: When `this.url` could be a falsy btw?

sidvishnoi, line 319: When `this.url` could be a falsy btw?

src/background/services/paymentSession.ts
```diff
@@ -300,4 +307,61 @@ export class PaymentSession {
...
+
+    const sessionIndex = overpayingSessions.findIndex(
+      (session) =>
+        session.tabId === this.tabId &&
+        session.hashUrl === hashUrl &&
+        session.walletAddress === this.receiver
+    )
```
sidvishnoi, line 326: Can we make has this entirely as the key in storage? Can make key something like `{tabId}:{hashUrl}:{hashReceiver}`

src/background/services/background.ts
sidvishnoi: Agree on not keeping it in primary extension storage though.
Session storage has added benefit of not needing to do https://github.com/interledger/web-monetization-extension/issues/287 maybe?

raducristianpopa: Can we specify a TTL for the items that are added in the session storage?

sidvishnoi: We can't, but we can have a guarantee that it goes away when browser is closed.
To save storage, we'd need to clear expired items ourselves (which we can do while setting a value, instead of requiring a new worker service).

src/background/services/paymentSession.ts
dianafulga: I'm using the url from the sender of browser.runtime, which is optional

raducristianpopa:

<table role="table">
<thead>
  <tr>
    <th>Name</th>
    <th>Link</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td align="center">9b4dc4c05ca7099f66427dd500b69c3ee12aef6c</td>
  </tr>
  <tr>
    <td align="center"><a href="https://github.com/interledger/web-monetization-extension/actions/runs/9592399714">Run #9592399714</a></td>
  </tr>
  <tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/25108303262/artifacts/1619453177">Download</a></td></tr><tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/25108303262/artifacts/1619453178">Download</a></td></tr>
</tbody>
</table>

src/background/services/background.ts
sidvishnoi: Just checked `storage.session` is available from:
- Chrome 102 (May 26, 2022)
- Safari 16.4 (March 27, 2023)
- Firefox 115 (July 4, 2023)

So, might need to use `storage.local` only for now.

src/background/utils.ts
```diff
@@ -84,8 +84,21 @@ export const getTabId = (sender: Runtime.MessageSender): number => {
...
   const tabId = getTabId(sender)
   const frameId = notNullOrUndef(sender.frameId, 'sender.frameId')
-  return { tabId, frameId }
+  return { tabId, frameId, url: sender.url }
 }
 
 export const computeRate = (rate: string, sessionsCount: number) =>
   (+rate / sessionsCount).toString()
+
+export const getHash = async (str: string) => {
```
sidvishnoi, line 93: Consider memoizing this.

src/background/services/paymentSession.ts
```diff
@@ -122,6 +127,8 @@ export class PaymentSession {
 
     let outgoingPayment: OutgoingPayment | undefined
 
+    this.tabState.processOverpaying(this.tab, this.url, this.receiver)
```
sidvishnoi, line 130:
```suggestion
    await this.tabState.processOverpaying(this.tab, this.url, this.receiver)
```

src/background/services/paymentSession.ts
```diff
@@ -172,6 +179,13 @@ export class PaymentSession {
             }
           })
 
+          this.tabState.saveOverpaying(
+            this.tab,
+            this.url,
+            this.receiver,
+            this.intervalInMs
```
raducristianpopa, line 186: Non overpaying sessions are getting saved as well. We should only save the overpaying ones.

src/background/services/tabState.ts
```diff
@@ -0,0 +1,64 @@
...
+import { sleep } from '@/shared/helpers'
+import { WalletAddress } from '@interledger/open-payments'
+
+type State = {
+  lastPaymentTimestamp: number
+  expireTimestamp: number
+}
+
+export class TabState {
+  private states = new WeakMap<Tabs.Tab, { [key: string]: State }>()
```
raducristianpopa, line 13:
```suggestion
  private state = new WeakMap<Tabs.Tab, { [key: string]: State }>()
```

src/background/services/tabState.ts
```diff
@@ -0,0 +1,64 @@
+import { getHash } from '../utils'
+
+import { Tabs } from 'webextension-polyfill'
+import { sleep } from '@/shared/helpers'
+import { WalletAddress } from '@interledger/open-payments'
+
+type State = {
+  lastPaymentTimestamp: number
+  expireTimestamp: number
```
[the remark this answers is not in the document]
dianafulga: I would still keep the timestamp word, to know why it's a number, not a date, and have the same for the latPayment. Maybe name it expiresAtTimestamp?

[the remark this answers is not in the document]
sidvishnoi: I think timestamp makes more sense with ISO string etc. Adding a comment would be fine in either case.

src/background/services/tabState.ts
```diff
@@ -0,0 +1,62 @@
...
+  private state = new WeakMap<Tabs.Tab, { [key: string]: State }>()
+
+  constructor() {}
+
+  private async getStateKey(
+    url: string,
+    walletAddress: WalletAddress
+  ): Promise<string> {
+    const hashUrl = await getHash(url)
+    const hashWalletAddress = await getHash(JSON.stringify(walletAddress))
```
sidvishnoi, line 21: Maybe we can get rid of hashing given we're storing in memory only?

src/background/services/monetization.ts
```diff
@@ -74,10 +82,13 @@ export class MonetizationService {
         receiver,
         connectedWallet,
         requestId,
+        tab,
         tabId,
         frameId,
```
sidvishnoi, line 87: Should we just pass the `sender` and get `tabId` `frameId`, `tab` in `PaymentSession`?

src/background/services/tabState.ts
sidvishnoi: Might be better to use `new WeakMap<Tabs.Tab, Map<string, State>>()` also.

sidvishnoi: Also, do we need to keep entire `walletAddress`? Can't we only use `walletAddress.id`?

sidvishnoi, line 21: Also, do we need to keep entire `walletAddress`? Can't we only use `walletAddress.id`?

src/background/services/tabState.ts
```diff
@@ -0,0 +1,62 @@
...
+    const key = `${hashUrl}:${hashWalletAddress}`
```
sidvishnoi, line 22: Another: do we need URL in here? Tab intrinsically has the URL, and we mostly care about not paying same walletAddress in same tab here.

src/background/services/tabState.ts
```diff
@@ -0,0 +1,62 @@
...
+  ): Promise<void> {
+    if (!intervalInMs) return
+
+    const key = await this.getStateKey(url, walletAddress)
+    const state = this.state.get(tab)?.[key]
+
+    const crtTimestamp = Date.now()
+    const expiresAtTimestamp = crtTimestamp + intervalInMs
+
+    if (state) {
```
sidvishnoi, line 57: Bug: We're not handling the case where `key` is to be set in state (i.e. `state` is never really set).

src/background/utils.ts
```diff
@@ -1,5 +1,5 @@
 import { WalletAmount } from '@/shared/types'
-import { type Browser, Runtime } from 'webextension-polyfill'
+import { type Browser, Runtime, Tabs } from 'webextension-polyfill'
```
sidvishnoi, line 2:
```suggestion
import type { Browser, Runtime, Tabs } from 'webextension-polyfill'
```

src/background/services/paymentSession.ts
```diff
@@ -122,6 +127,12 @@ export class PaymentSession {
 
     let outgoingPayment: OutgoingPayment | undefined
 
+    const waitTime = await this.tabState.getOverpayingWaitTime(this.tab, this.url, this.receiver)
+
+    if (waitTime) {
+      await sleep(waitTime)
+    }
+
```
sidvishnoi, line 135: Nit: If we always return `number` (0 in other case), we just `await sleep(0)`. Less things to think of (less branches).

src/background/services/paymentSession.ts
```diff
@@ -172,6 +183,15 @@ export class PaymentSession {
             }
           })
 
+          if (this.intervalInMs > 1000) {
+            this.tabState.saveOverpaying(
```
sidvishnoi, line 187: I'm not following how we define "overpaying". We can just store that state always? It's ok if `waitTime` is 0. That data can be stored bookkeeping only (in memory only).

src/background/services/tabState.ts
dianafulga: I'd keep the hash  for consistency (in the feature we'll implement the exception list, that will also use the url, better have the hash everywhere)

src/background/services/monetization.ts
dianafulga: we need to call getSender here anyway, calling it again in the PaymentSession seemed to much

src/background/services/tabState.ts
sidvishnoi: But exception list will be in storage (privacy part), and not updated on each payment. Speaking from performance perspective only - our extension will consume more battery.

src/background/services/tabState.ts
dianafulga: we can have the same walletAddress in 2 different iframes, so we need the url

src/background/services/paymentSession.ts
dianafulga: If we are paying every second, there is no use of storing all the data.  cc: @raducristianpopa

src/background/services/tabState.ts
sidvishnoi: So, if a site adds 10 iframes (with different URLs) using same wallet address on a single page, that won't count as overpaying?

src/background/services/paymentSession.ts
dianafulga: from what I know sleep 0 will still put the promise in the queue and it's picked up immediately after. I will keep it how we have it

sidvishnoi: Which is fine. Very little impact on performance. It'll get optimized by browser.
I prefer to make code simpler to readers - so less decisions in mind, the better. A nitpick anyway, feel free to ignore

src/background/services/tabState.ts
dianafulga: I would say no. We can discuss with the team and treat case this in another task

src/background/services/tabState.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  private getStateKey(url: string, walletAddressId: string): string {
+    return `${url}:${walletAddressId}`
+  }
+
+  async getOverpayingWaitTime(
+    tab: Tabs.Tab,
+    url: string,
+    walletAddressId: string
+  ): Promise<number> {
```
sidvishnoi, line 21:
```suggestion
  getOverpayingWaitTime(
    tab: Tabs.Tab,
    url: string,
    walletAddressId: string
  ): number {
```

src/background/services/tabState.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  async saveOverpaying(
+    tab: Tabs.Tab,
+    url: string,
+    walletAddressId: string,
+    intervalInMs: number
+  ): Promise<void> {
+    if (!intervalInMs) return
+
+    const crtTimestamp = Date.now()
```
sidvishnoi, line 41:
```suggestion
    const now = Date.now()
```

```suggestion
    const currentTimestamp = Date.now()
```

src/background/services/tabState.ts
```diff
@@ -0,0 +1,62 @@
...
+      const tabState = this.state.get(tab) || new Map()
+      tabState.set(key, state)
+
+      this.state.set(tab, tabState)
+    }
+
+    if (state) {
+      state.expiresAtTimestamp = expiresAtTimestamp
+      state.lastPaymentTimestamp = crtTimestamp
+    }
```
sidvishnoi, line 60:
```suggestion
    state.expiresAtTimestamp = expiresAtTimestamp
    state.lastPaymentTimestamp = crtTimestamp
```

src/background/services/tabState.ts
```diff
@@ -0,0 +1,62 @@
...
+    const crtTimestamp = Date.now()
+    const expiresAtTimestamp = crtTimestamp + intervalInMs
+
+    const key = this.getStateKey(url, walletAddressId)
+    const state = this.state.get(tab)?.get(key) || {
+      expiresAtTimestamp: expiresAtTimestamp,
+      lastPaymentTimestamp: crtTimestamp
+    }
+
+    if (!state) {
```
sidvishnoi, line 50: Bug: `state` is always set above.

src/background/utils.ts
```diff
@@ -81,11 +81,30 @@ export const getTabId = (sender: Runtime.MessageSender): number => {
...
+  const hash = await crypto.subtle.digest('SHA-1', encoder.encode(str))
+
+  return Array.from(new Uint8Array(hash))
+    .map((v) => v.toString(16).padStart(2, '0'))
+    .join('')
+}
+
+export const removeQueryParams = (url: string) => {
+  return url.split('?')[0]
+}
```
raducristianpopa, line 110: Let's use the `URL` constructor for this.

src/background/services/tabState.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  async saveOverpaying(
+    tab: Tabs.Tab,
+    url: string,
+    walletAddressId: string,
+    intervalInMs: number
+  ): Promise<void> {
+    if (!intervalInMs) return
+
+    const currentTimestamp = Date.now()
```
raducristianpopa, line 41: Consistency:
```suggestion
    const now = Date.now()
```

src/background/services/paymentSession.ts
```diff
@@ -172,6 +185,15 @@ export class PaymentSession {
             }
           })
 
+          if (this.intervalInMs > 1000) {
```
raducristianpopa, line 188: Can you add a to do item here: I feel like relying on the interval isn't really a source of truth. We can have a flag in the session that we set when we call the `adjustSessionAmount` method.

dianafulga: Done. But based on what will we compute the flag? We can implement it in this one https://github.com/interledger/web-monetization-extension/issues/375

src/background/services/tabState.ts
```diff
@@ -0,0 +1,60 @@
...
+  lastPaymentTimestamp: number
+  expiresAtTimestamp: number
+}
+
+export class TabState {
+  private state = new WeakMap<Tabs.Tab, Map<string, State>>()
+
+  constructor() {}
+
+  private getStateKey(url: string, walletAddressId: string): string {
```
sidvishnoi, line 13: Keeping https://github.com/interledger/web-monetization-extension/issues/381 in mind, maybe should change current `this.state` to something like `this.lastPaymentInfoState` (as the keys could be different)?

```suggestion
  private getOverpayingStateKey(url: string, walletAddressId: string): string {
```

src/background/services/tabState.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    return 0
+  }
+
+  async saveOverpaying(
+    tab: Tabs.Tab,
+    url: string,
+    walletAddressId: string,
+    intervalInMs: number
+  ): Promise<void> {
```
sidvishnoi, line 38:
```suggestion
  saveOverpaying(
    tab: Tabs.Tab,
    url: string,
    walletAddressId: string,
    intervalInMs: number
  ): void {
```

src/background/services/tabState.ts
```diff
@@ -0,0 +1,60 @@
...
+    const now = Date.now()
+    const expiresAtTimestamp = now + intervalInMs
+
+    const key = this.getStateKey(url, walletAddressId)
+    const state = this.state.get(tab)?.get(key) || {
+      expiresAtTimestamp: expiresAtTimestamp,
+      lastPaymentTimestamp: now
+    }
+
+    if (!state) {
```
sidvishnoi, line 50: Bug: state is always as set above. Perhaps use something like:

```ts
let state =  this.state.get(tab)?.get(key)
if (!state) {
  const tabState = this.state.get(tab) || this.state.set(tab, new Map()).get(tab)!
  tabState.set(key, { lastPaymentTimestamp: now, ... })
} else {
  state.lastPaymentTimestamp = now
  // ...
}

File: src/background/utils.ts [hunk 26]
```diff
@@ -81,11 +81,37 @@ export const getTabId = (sender: Runtime.MessageSender): number => {
...
+
+export const removeQueryParams = (urlString: string) => {
+  const url = new URL(urlString)
+
+  const origin = url.origin
+  const pathname = url.pathname
+
+  const fullPath = origin + pathname
+
+  return fullPath
```
sidvishnoi, line 116:
```suggestion
  return url.origin + url.pathname
```

src/background/utils.ts
```diff
@@ -81,11 +81,37 @@ export const getTabId = (sender: Runtime.MessageSender): number => {
...
-  return { tabId, frameId }
+  const tab = getTab(sender)
+
+  return { tabId, frameId, url: sender.url, tab }
 }
 
 export const computeRate = (rate: string, sessionsCount: number) =>
   (+rate / sessionsCount).toString()
+
+export const getHash = async (str: string) => {
```
sidvishnoi, line 99: We can remove this function as unused, or feel free to keep it. I'd recommend passing algorithm and format to make it more usable (can use in `OpenPayments.verifyInteractionHash`).. or maybe something for future only..

```ts
export const getHash = async (str: string, algorithm: AlgorithmIdentifier, format: 'hex' | 'base64') => {
```

src/shared/types.ts
```diff
@@ -27,7 +27,6 @@ export interface GrantDetails {
   accessToken: string
   continueUri: string
 }
-
```
sidvishnoi, line 30: Nit: undo this deleted line, as that can make git logs confusing.

sidvishnoi (changes requested):
Mainly that state-always-defined bug need resolved.

sidvishnoi (approved):

Issue #332, opened 2024-06-13
[BUG] Extension unable to disconnect after  public key has already been dissociated from the wallet 


RabebOthmani:
### Steps to reproduce

1. Go to rafiki.money --> Settings --> Developer Keys 
2. Revoke key from wallet 
3. On the extension UI, click on the disconnect button 
4. Extension unable to disconnect 

### Expected result

Extension should gracefully disconnect from Wallet 

### Actual result

Extension shows payment screen immediately after clicking on disconnect. 
Going back to the extension settings screen shows wallet still connected 

### Operating system

Windows

### Browsers

Chrome

### Extension version

Alpha

raducristianpopa:
For the disconnecting part, we can check if the returned error message is `invalid_client` when making the API call. If it is returning `invalid_client` can we make the assumption that the user does not have the key anymore and we can safely clear the storage`?

if this is happening during the monetization flow, should we have an error state for the extension? We can open the popup pragmatically when this happens. Maybe the error page in the popup can prompt the public key for the user again to be reuploaded, alongside a disconnect button that just clears the storage?

CC @sidvishnoi @dianafulga @tselit

sidvishnoi:
Sounds good.

We should also update the error message on connect screen when key isn't synced with wallet (instead of bad request one).

sidvishnoi:
For the action-required error states, I think we can add routes under /errors/*. This we we can keep the current default screens same, instead of adding alerts in them (which end up adding more scrollbars). I've that MissingHostPermission route, can merge there.
Ideally, we can add a ErrorLayout and "slot" content inside it for each error page (See MissingHostPermission one for example).

tselit:
**To do**: @tselit to update the Figma design file

sidvishnoi:
> For the disconnecting part, we can check if the returned error message is invalid_client when making the API call. If it is returning invalid_client can we make the assumption that the user does not have the key anymore and we can safely clear the storage`?

Just realized the client doesn't return any response of a bad request, so we can't really know if it's a `invalid_client` error or something else 😬. I think this is something that should be handled at client level, or in Rafiki - so we get better error message in return.

sidvishnoi:
Alternatively,
- we can assume 400 bad request in given state means key removed,
- or, use our own OpenPayments client, so we've better control.

raducristianpopa:
> Alternatively,
> 
>   we can assume 400 bad request in given state means key removed,
> 
[...]

Issue #147, opened 2024-03-07
Error handling/messages improvements


raducristianpopa:
### Context

The extension UI error handling can be improved by displaying more meaningful error messages to the user.

### Todos

- [x] Audit popup for better error messages
- [x] Audit background response the popup for better error messages
- [x] ~What to display when a grant is out of funds~ https://github.com/interledger/web-monetization-extension/issues/144

sidvishnoi:
Most likely also blocked by https://github.com/interledger/open-payments/issues/482.

sidvishnoi:
Looks much better now. Address through multiple PRs.

PR #366, opened 2024-06-25
fix(background): handle key removed from wallet

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/332.
Part of https://github.com/interledger/web-monetization-extension/issues/147.

1. Add `key_revoked` ExtensionState.
2. Catch "key revoked" errors OpenPayments service when making payments, and switch to `key_revoked` state.
3. Allow disconnect if key revoked.
4. Add `ErrorKeyRevoked` screen in popup, made of two screens:
   - `MainScreen`: lets user choose if they want to disconnect or reconnect
   - `ReconnectScreen`: visible after user chooses they want to reconnect
5. Update Monetization service to:
   - Stop all sessions on key revoke (as soon as `open_payments.key_revoked` event is fired from a PaymentSession)
   - Update behavior:
      - Before resuming/starting any session, check for `key_revoked` state (along with connected/enabled)
      - But allow stopping payments regardless of above state.
6. Move some utils to shared directory, add some unit tests for them.

> [!NOTE]
> - Disconnect button on `MainScreen` leads to direct disconnect and switch to default setup screen.
> - Once user choose they want to reconnect, that screen state is stored in localStorage (with newly added `useLocalStorage` helper), so when they've copied the key, they come back to `ReconnectScreen` only. That state is stored for 2min only.

> [!WARNING]
> There's no way to go back to MainScreen from ReconnectScreen as of now. Need to integrate that icon in header somehow (https://github.com/interledger/web-monetization-extension/issues/424). Or let it be, as the Settings screen has disconnect button anyway.


raducristianpopa:

<table role="table">
<thead>
  <tr>
    <th>Name</th>
    <th>Link</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td align="center">d67bc17083fdf20690a4c21c94a34a19cf6285f5</td>
  </tr>
  <tr>
    <td align="center"><a href="https://github.com/interledger/web-monetization-extension/actions/runs/9660944604">Run #9660944604</a></td>
  </tr>
  <tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/25285014808/artifacts/1635492555">Download</a></td></tr><tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/25285014808/artifacts/1635492557">Download</a></td></tr>
</tbody>
</table>

src/background/services/monetization.ts
```diff
@@ -170,19 +174,19 @@ export class MonetizationService {
...
       return
     }
 
+    const { state, connected, enabled } = await this.storage.get([
+      'state',
+      'connected',
+      'enabled'
+    ])
+    if (state !== null || !connected || !enabled) return
+
```
sidvishnoi, line 189: TODO: refactor this into `canMakePayment` something. Also, for better performance, store state, connected and enabled as member fields, and not access from storage each time (making sure they're synced of course).

src/background/services/openPayments.ts
```diff
@@ -501,13 +501,23 @@ export class OpenPaymentsService {
...
-      this.token = { value: '', manageUrl: '' }
+    if (!grant) {
+      return
     }
+
+    await this.client!.grant.cancel(grant.continue).catch((error) => {
+      if (error instanceof OpenPaymentsClientError) {
+        if (error.status === 400 && error.code === 'invalid_client') {
+          // key already removed from wallet
+          return
```
sidvishnoi, line 512: Nit: went with `.catch` instead of `try {} catch {}` as can return easily here.

src/popup/components/ui/Button.tsx
```diff
@@ -40,7 +40,8 @@ export interface ButtonProps
   extends VariantProps<typeof buttonVariants>,
     React.ButtonHTMLAttributes<HTMLButtonElement> {
   loading?: boolean
-  ['aria-label']: string
+  /** Optional only when children are passed */
+  ['aria-label']?: string
```
sidvishnoi, line 44: Setting `aria-label` while passing textContent to Button is counter-productive.

sidvishnoi:
Demo (sorry for lack in cursor in recording):

[Screencast from 08-07-24 06:41:37 PM IST.webm](https://github.com/interledger/web-monetization-extension/assets/8426945/ba0cc745-ac78-453a-ab91-686cced74035)

src/popup/lib/utils.ts
```diff
@@ -62,3 +64,34 @@ export function formatNumber(
     } else return value.toExponential()
   }
 }
+
+export function useLocalStorage<T>(key: string, defaultValue?: T) {
```
sidvishnoi, line 68: I'll add an expires duration here, for safety.

src/popup/lib/utils.ts
```diff
@@ -62,3 +64,59 @@ export function formatNumber(
     } else return value.toExponential()
   }
 }
+
+export function useLocalStorage<T>(
```
sidvishnoi, line 68: If this util is judged useful, I can move this into a separate PR.

src/background/services/monetization.ts
```diff
@@ -259,6 +265,20 @@ export class MonetizationService {
...
   }
 
+  private onKeyRevoked() {
+    this.events.once('open_payments.key_revoked', async () => {
+      this.logger.warn(`Key revoked. Stopping all payment sessions.`)
+      for (const sessions of Object.values(this.sessions)) {
+        for (const session of sessions.values()) {
+          session.stop()
+        }
+      }
```
sidvishnoi, line 275: Do we want to stop all or just active tab(s) ones?
My concern here is how to works in Edge's split screen view.

src/_locales/en/messages.json
```diff
@@ -24,6 +24,21 @@
...
+  },
+  "error_keyRevoked_text": {
+    "message": "It appears that the key has been revoked from your wallet. Authenticate wallet by adding the key again, or disconnect."
+  },
+  "action_keyRevoked_disconnect": {
+    "message": "Yes, let me disconnect my wallet"
+  },
+  "action_keyRevoked_reconnect": {
+    "message": "It was a mistake, I’d like to reconnect my wallet"
+  },
```
raducristianpopa, line 38: I feel like we are being to wordy for these two. Thoughts on:
```suggestion
  "action_keyRevoked_disconnect": {
    "message": "Disconnect my wallet"
  },
  "action_keyRevoked_reconnect": {
    "message": "Reconnect my wallet"
  },
```
CC @RabebOthmani

src/background/services/openPayments.ts
raducristianpopa: In this case, if the key is revoked and the user wants to disconnect their wallet we should clear the storage. The user can revoke the grant in the wallet UI as well.

src/background/services/monetization.ts
raducristianpopa: Hmm, thanks for pointing out Edge's split screen view. I'm okay with the implementation as is, but we need to spend some quality time on Edge as well to check how this would work. Let's create an issue for this.

src/background/services/paymentSession.ts
```diff
@@ -204,8 +206,16 @@ export class PaymentSession {
...
+    try {
+      const incomingPayment = await this.createIncomingPayment()
+      this.incomingPaymentUrl = incomingPayment.id
+    } catch (error) {
+      if (this.isKeyRevokedError(error)) {
+        this.events.emit('open_payments.key_revoked')
+        return
+      }
+      throw error
+    }
```
raducristianpopa, line 218: 👍

src/popup/lib/utils.ts
raducristianpopa: Yes. Let's move this in a separate PR.

src/popup/lib/utils.ts
```diff
@@ -62,3 +64,59 @@ export function formatNumber(
...
+    maxAge = Infinity
+  }: Partial<{ defaultValue: T; maxAge: number }> = {}
+) {
+  const hasLocalStorage = typeof localStorage !== 'undefined'
+  type Stored = { value: T; ts: number }
+  const isWellFormed = (obj: any): obj is Stored => {
+    if (typeof obj !== 'object' || obj == null) return false
+    if (!obj.ts || !Number.isSafeInteger(obj.ts)) return false
+    return typeof obj.value !== 'undefined'
+  }
```
raducristianpopa, line 81: We will probably not have a lot of re-renders, but I think we can wrap this in a `useCallback`.

src/popup/lib/utils.ts
```diff
@@ -62,3 +64,59 @@ export function formatNumber(
...
+
+export function useLocalStorage<T>(
+  key: string,
+  {
+    defaultValue,
+    maxAge = Infinity
+  }: Partial<{ defaultValue: T; maxAge: number }> = {}
+) {
+  const hasLocalStorage = typeof localStorage !== 'undefined'
+  type Stored = { value: T; ts: number }
```
raducristianpopa, line 76: Let's move the type outside of the function block.

src/popup/lib/utils.ts
raducristianpopa, line 1: I would move the `useLocalStorage` hook to a new file: `lib/hooks.ts`.

raducristianpopa (commented):
Only reviewed the code - I still need to test the new functionalities.

src/background/services/monetization.ts
raducristianpopa: Add a todo comment or log an issue for this, please. 🙏

src/popup/lib/utils.ts
sidvishnoi: It's very much internal to this function - never exposed outside. So, I'd prefer to keep this one in.

src/background/services/openPayments.ts
sidvishnoi: Which is what it's doing here? To be clear, we're ignoring the error only with this return here - it's not returning from `disconnectWallet()` here, so we still clear the storage.

src/background/services/monetization.ts
sidvishnoi: This loop would need to be modified regardless in https://github.com/interledger/web-monetization-extension/issues/381 if we store tabs in a WeakMap (Weakmaps can't be iterated, so we'd query active tabs and then iterate over that).

src/background/services/openPayments.ts
raducristianpopa: My bad, I did not notice `.catch()`.

src/popup/lib/utils.ts
sidvishnoi: Sent https://github.com/interledger/web-monetization-extension/pull/419

src/_locales/en/messages.json
sidvishnoi: For reference: [this part in Figma](https://www.figma.com/design/0S8Oj8ZB5MvkGtIrVRswDH/Web-monetization?node-id=474-413&m=dev)

src/_locales/en/messages.json
```diff
@@ -24,19 +24,19 @@
...
-  "action_keyRevoked_disconnect": {
+  "keyRevoked_action_disconnect": {
     "message": "Yes, let me disconnect my wallet"
   },
-  "action_keyRevoked_reconnect": {
+  "keyRevoked_action_reconnect": {
     "message": "It was a mistake, I’d like to reconnect my wallet"
   },
-  "action_keyRevoked_reconnectBtn": {
+  "keyRevoked_action_reconnectBtn": {
```
raducristianpopa, line 39: We should to the same for the other items as well.

src/popup/components/ErrorKeyRevoked.tsx
```diff
@@ -0,0 +1,192 @@
...
+    setIsLoading(false)
+  }
+
+  return (
+    <m.div
+      initial={{ opacity: 0 }}
+      animate={{ opacity: 1 }}
+      exit={{ opacity: 0 }}
+      className="space-y-4 text-sm"
+    >
```
raducristianpopa, line 94: Not sure if it's only me, but I find this a little bit too flashy? I'm okay with keeping the animation, but only when switching between states and remove the animation on first render.

src/popup/components/ErrorKeyRevoked.tsx
```diff
@@ -0,0 +1,192 @@
...
+      exit={{ opacity: 0 }}
+      className="flex flex-col items-stretch gap-4"
+      onSubmit={handleSubmit(requestReconnect)}
+    >
+      <div className="space-y-1 text-xs">
+        <p className="px-2">Reconnecting to wallet: {info.walletAddress?.id}</p>
+        <p className="px-2">
+          <strong>Before</strong> you reconnect, copy the public key below and
+          add it to your wallet.
+        </p>
```
raducristianpopa, line 172: Should these be added to `_locales`?

src/popup/components/ErrorKeyRevoked.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  return (
+    <m.form
+      initial={{ opacity: 0 }}
+      animate={{ opacity: 1 }}
+      exit={{ opacity: 0 }}
+      className="flex flex-col items-stretch gap-4"
+      onSubmit={handleSubmit(requestReconnect)}
+    >
+      <div className="space-y-1 text-xs">
```
raducristianpopa, line 167: The text is so small. Let's make it bigger.

src/popup/components/ErrorKeyRevoked.tsx
```diff
@@ -0,0 +1,192 @@
...
+        <p className="px-2">Reconnecting to wallet: {info.walletAddress?.id}</p>
```
raducristianpopa, line 168: Should we put some accent on the wallet address (bold, another color)?

raducristianpopa (commented):
Tested and reviewed the UI part. Left a few a comments.

I know we don't have this in the design, but I feel like we should add a back button on the reconnect screen - going back to the initial page where the user can choose between reconnecting and disconnecting.

While in the `key_revoked` state, we should disable the settings button from the top bar - at the moment the user can switch to the settings page when in this state.

src/_locales/en/messages.json
sidvishnoi: Was avoiding changing other stuff as PR is already large enough. Will send a follow-up PR updating all other keys.

src/popup/components/ErrorKeyRevoked.tsx
sidvishnoi: I wanted to, but that `<strong>` stopped me. We would have to use `dangerouslySetInnerHTML`, which I was avoiding.

src/popup/components/ErrorKeyRevoked.tsx
sidvishnoi: Heh. Copied from Figma. Will do.

src/popup/components/ErrorKeyRevoked.tsx
sidvishnoi: I think underline could be nice?

sidvishnoi:
> but I feel like we should add a back button on the reconnect screen - going back to the initial page where the user can choose between reconnecting and disconnecting.

That will require some major message passing to parent, as header belongs to `MainLayout`. I think we can use something with `useContext`, but it'll require some thinking, so skipped for now.

raducristianpopa:
> > but I feel like we should add a back button on the reconnect screen - going back to the initial page where the user can choose between reconnecting and disconnecting.
> 
> That will require some major message passing to parent, as header belongs to `MainLayout`. I think we can use something with `useContext`, but it'll require some thinking, so skipped for now.

Here we are talking about the back button or about disabling the settings one? Got confused by the quote.

sidvishnoi:
> > > but I feel like we should add a back button on the reconnect screen - going back to the initial page where the user can choose between reconnecting and disconnecting.
> > 
> > 
> > That will require some major message passing to parent, as header belongs to `MainLayout`. I think we can use something with `useContext`, but it'll require some thinking, so skipped for now.
> 
> Here we are talking about the back button or about disabling the settings one? Got confused by the quote.

I kept settings button on as there's no other way to disconnect after selecting reconnect (except waiting two minutes and opening popup again). Design suggests we'd show the back-arrow after user selects Reconnect, to go back to previous screen.

raducristianpopa (approved):

PR #367, opened 2024-06-25
fix(content): dispatch `load` event if wallet address is valid

### LINKED ISSUE #365: [BUG] Load event not being triggered
### Steps to reproduce

1. Have a connected wallet
2. Go to a web monetized website that has a listener for the `load` event on the monetization tag
3. If the wallet address is valid, the `load` event is not being dispatched

### Expected result

The `load` event should be triggered if the tag href is a valid wallet address

### Actual result

`load` event no being dispatched.

### Screenshots or videos

_No response_

### Additional context

_No response_

### Operating system

macOS

### Operating system version

_No response_

### Browsers

Chrome, Microsoft Edge, Firefox, Opera

### Browser version

_No response_

### Extension version

alpha 2

raducristianpopa:
## Context

Closes #365.

Initially I thought there was an error and the `load` event was not correctly dispatched. It looks like we omitted this. Now we dispatch the `load` event if the wallet address is valid.

Additional changes:
- small refactoring for the wallet address check
- update the `checkWalletAddressUrlFormat` function since we do not need to check for SPSP endpoints

### DIFF
--- a/src/content/services/monetizationTagManager.ts
+++ b/src/content/services/monetizationTagManager.ts
@@ -404,6 +404,8 @@ export class MonetizationTagManager extends EventEmitter {
     crtRequestId?: string
   ): Promise<StartMonetizationPayload | null> {
     const walletAddress = await this.checkTag(tag)
+    if(!walletAddress) return null
+
     const requestId = crtRequestId ?? crypto.randomUUID()
     const details: MonetizationTagDetails = {
       walletAddress,
@@ -411,7 +413,7 @@ export class MonetizationTagManager extends EventEmitter {
     }
 
     this.monetizationTags.set(tag, details)
-    return walletAddress ? { walletAddress, requestId } : null
+    return { walletAddress, requestId }
   }
 
   private sendStartMonetization(tags: StartMonetizationPayload[]) {
@@ -496,49 +498,28 @@ export class MonetizationTagManager extends EventEmitter {
 
     if (tag.hasAttribute('disabled')) return null
 
-    const isValidWalletAddressUrlFormat = this.checkWalletAddressUrlFormat(tag)
+    const walletAddressInfo = await this.validateWalletAddress(tag)
 
-    return isValidWalletAddressUrlFormat
-      ? await this.checkWalletAddressUrlCall(tag)
-      : null
+    return walletAddressInfo
   }
 
-  // Check wallet address url for valid format
-  private checkWalletAddressUrlFormat(tag: MonetizationTag) {
-    const walletAddressUrl = tag.href.trim()
+  private async validateWalletAddress(tag: MonetizationTag): Promise<WalletAddress | null>{
+      const walletAddressUrl = tag.href.trim()
+      try {
+        checkWalletAddressUrlFormat(walletAddressUrl)
+        const response = await checkWalletAddressUrlCall({walletAddressUrl})
 
-    try {
-      checkWalletAddressUrlFormat(walletAddressUrl, true)
+        if(response.success === false) {
+            throw new Error(`Could not retrieve wallet address information for ${JSON.stringify(walletAddressUrl)}.`)
+        }
 
-      return true
-    } catch (err) {
-      this.logger.error(err)
-
-      const event = new Event('error')
-      tag.dispatchEvent(event)
-
-      return false
-    }
-  }
-
-  // Check wallet address url by fetching the wallet address details
-  private async checkWalletAddressUrlCall(
-    tag: MonetizationTag
-  ): Promise<WalletAddress | null> {
-    const walletAddressUrl = tag.href.trim()
-
-    try {
-      const response = await checkWalletAddressUrlCall({ walletAddressUrl })
-
-      return response.success ? response.payload : null
-    } catch (err) {
-      this.logger.error(err)
-
-      const event = new Event('error')
-      tag.dispatchEvent(event)
-
-      return null
-    }
+        tag.dispatchEvent(new Event('load'))
+        return response.payload
+      } catch(e) {
+        this.logger.error(e)
+        tag.dispatchEvent(new Event('error'))
+        return null
+      }
   }
 
   private bindMessageHandler() {
--- a/src/content/services/monetizationTagManager.ts
+++ b/src/content/services/monetizationTagManager.ts
@@ -404,7 +404,7 @@ export class MonetizationTagManager extends EventEmitter {
     crtRequestId?: string
   ): Promise<StartMonetizationPayload | null> {
     const walletAddress = await this.checkTag(tag)
-    if(!walletAddress) return null
+    if (!walletAddress) return null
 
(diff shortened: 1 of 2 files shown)


raducristianpopa:

<table role="table">
<thead>
  <tr>
    <th>Name</th>
    <th>Link</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td align="center">1b47918ca607a6141c63003e7657262a8f51b2c0</td>
  </tr>
  <tr>
    <td align="center"><a href="https://github.com/interledger/web-monetization-extension/actions/runs/9661173418">Run #9661173418</a></td>
  </tr>
  <tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/25285678025/artifacts/1635554562">Download</a></td></tr><tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/25285678025/artifacts/1635554563">Download</a></td></tr>
</tbody>
</table>

src/content/utils.ts
```diff
@@ -1,55 +1,32 @@
...
-        JSON.stringify({
-          hash: url.hash,
-          search: url.search,
-          port: url.port,
-          username: url.username,
-          password: url.password
-        })
+  if (hash || search || port || username || password) {
+    throw new WalletAddressFormatError(
+      `Wallet address URL must not contain query/fragment/port/username/password elements. Received: ${JSON.stringify({ hash, search, port, username, password })}`
```
sidvishnoi, line 27: Perhaps we can do a comparison with what we allow, instead of showing the technical hash, search etc parts? 
For example: `if (urlWeAreOkWith !== UrlThatUserGave) throw UserGaveExtraStuff`

src/content/utils.ts
```diff
@@ -1,55 +1,32 @@
...
   try {
-    url = new URL(httpUrl)
-    if (!url.protocol.match(/https?:/)) {
-      const spec = urlOnly ? '' : 'either a payment pointer or '
-      // noinspection ExceptionCaughtLocallyJS
-      throw new PaymentEndpointError(
-        `SPSP endpoint must be specified as ${spec}fully resolved https:// url, ` +
-          `got ${JSON.stringify(pointerOrUrl)} `
+    url = new URL(walletAddressUrl)
+    if (url.protocol !== 'https:') {
```
sidvishnoi, line 7:
```suggestion
    if (!ALLOWED_PROTOCOLS.includes(url.protocol)) {
```

src/content/utils.ts
sidvishnoi: or, we can just accept and ignore rest of the parts (when saving) if we get the same result anyway?

raducristianpopa: > Perhaps we can do a comparison with what we allow, instead of showing the technical hash, search etc parts?
For example: if (urlWeAreOkWith !== UrlThatUserGave) throw UserGaveExtraStuff

Would attaching a link to the Open Payments documentation work?

> or, we can just accept and ignore rest of the parts (when saving) if we get the same result anyway?

I think we should be strict about the format to not cause misconceptions.

src/content/utils.ts
raducristianpopa: The wallet address should always be served under `https:`.

sidvishnoi: I meant if in dev we want to use a localhost Rafiki.

src/content/utils.ts
sidvishnoi: We can edit the input to what we want and warn the user ("heads up, you pasted this but wallet address is just the URL that looks like this"); instead of adding friction to this user input.
IMO, we should even let input `$paymentPointer` (as can be copied from rafiki.money at present), to reduce this friction.

I think adding link to documentation is still technical enough for users, as technical as showing the `{query,fragment,port}` etc.

sidvishnoi: Let's do that as a follow-up... reducing friction. For now, let's use an error message that gives most info.

sidvishnoi: Ah! Never mind! I thought it was the input in popup!

src/content/services/monetizationTagManager.ts
```diff
@@ -496,47 +506,30 @@ export class MonetizationTagManager extends EventEmitter {
...
+        throw new Error(
+          `Could not retrieve wallet address information for ${JSON.stringify(walletAddressUrl)}.`
+        )
+      }
 
+      this.dispatchLoadEvent(tag)
+      return response.payload
+    } catch (e) {
+      this.logger.error(e)
+      this.dispatchErrorEvent(tag)
```
sidvishnoi, line 532: Should include error message in this (can dispatch `new ErrorEvent('error', {message})`), instead of logging from content script (optional instead, ok to show error twice I think).

sidvishnoi: Looks like ErrorEvent can't be transmitted across content script boundary. Using CustomElement like we do for `monetization` event should work.

```diff
diff --git a/src/content/polyfill.ts b/src/content/polyfill.ts
index acd8c65..7c03428 100644
--- a/src/content/polyfill.ts
+++ b/src/content/polyfill.ts
@@ -122,4 +122,19 @@ import type { MonetizationEventPayload } from '@/shared/messages'
     },
     { capture: true }
   )
+
+  window.addEventListener(
+    '__wm_ext_error',
+    (event: CustomEvent<{ message: string; error: Error }>) => {
+      if (!event.target) return
+      if (!(event.target instanceof HTMLElement)) return
+      event.target.dispatchEvent(
+        new ErrorEvent('error', {
+          message: event.detail.message,
+          error: event.detail.error
+        })
+      )
+    },
+    { capture: true }
+  )
 })()
diff --git a/src/content/services/monetizationTagManager.ts b/src/content/services/monetizationTagManager.ts
index 8105c50..b5e8a8c 100644
--- a/src/content/services/monetizationTagManager.ts
+++ b/src/content/services/monetizationTagManager.ts
@@ -69,8 +69,13 @@ export class MonetizationTagManager extends EventEmitter {
     tag.dispatchEvent(new Event('load'))
   }
 
-  private dispatchErrorEvent(tag: MonetizationTag) {
-    tag.dispatchEvent(new Event('error'))
+  private dispatchErrorEvent(tag: MonetizationTag, error: Error) {
+    tag.dispatchEvent(
+      new CustomEvent('__wm_ext_error', {
+        detail: mozClone({ message: error.message, error }, this.document),
+        bubbles: true
+      })
+    )
   }
 
   dispatchMonetizationEvent({ requestId, detail }: MonetizationEventPayload) {
@@ -529,7 +534,7 @@ export class MonetizationTagManager extends EventEmitter {
       return response.payload
     } catch (e) {
       this.logger.error(e)
-      this.dispatchErrorEvent(tag)
+      this.dispatchErrorEvent(tag, e as Error)
       return null
     }
   }
```

Issue #376, opened 2024-06-27
Update storage structure to support one-time + recurring grants; balance for each


sidvishnoi:
```ts
{
  recurringGrantRemainingBalance: "1",
  recurringGrant: {
    interval: "R/....", // previously amount.interval
    approvedAmount: "1000", // previously amount.value
    continue: { // previously storage.grant
      url: "https://..",
      accessToken: "...",
    },
    accessToken: { // previously storage.token
      value: "ADAS",
      manageUrl: "https://...",
    },
  },

  oneTimeGrantRemainingBalance: "342",
  oneTimeGrant: {
    // same as recurringGrant above except interval
  },
}
```

- Part of https://github.com/interledger/web-monetization-extension/issues/272
- Part of https://github.com/interledger/web-monetization-extension/issues/144

Issue #272, opened 2024-05-17
Show remaining balance of authorized amnt


tselit:
When the WM extension is connected to a digital wallet, there is an amount of money/value that the user authorizes the extension to access from the user's connected wallet. 

The extension pop-up should accurately update the displayed remaining balance of the authorized total value.

- [x] https://github.com/interledger/web-monetization-extension/issues/376
- [ ] https://github.com/interledger/web-monetization-extension/issues/378
- [ ] https://github.com/interledger/web-monetization-extension/pull/386

sidvishnoi:
Relevant: https://github.com/interledger/open-payments/issues/469

RabebOthmani:
Is this ticket still blocked @raducristianpopa @sidvishnoi?

sidvishnoi:
Correct. Still blocked.

RabebOthmani:
What's blocking it? Tseli mentioned that the Open Payments issue is resolved. I guess we also need to define the UI

sidvishnoi:
Pending on releases (Rafiki, Open Payments client).

UI: I think there's something in the Figma (link in `#web-monetization-extension` Slack channel's pinned message), but not sure if it's outdated

sidvishnoi:
https://rafiki.money now returns `grantSpentDebitAmount`.
Now blocked only on open-payments SDK update.

Issue #144, opened 2024-03-07
Handle scenarios when budget exceeded


raducristianpopa:
## User has a one time amount

If we reached the grant limit - we cannot create outgoing payments anymore, we display a screen in the popup (inactive icon + orange dot - user interaction) where the user will have the options to:
- Set a new one time amount: we create a new grant for the amount that the user is going to specify and revoke the previous grant
- Set a recurring amount: we create a new grant for the amount that the user is going to specify with a recurring interval (30d) and revoke the previous grant
When the user runs out of funds we need to update the extension state (new property in the storage - out_of_funds

## User has a recurring amount

If we reached the grant limit, we display the same screen as in the one time amount scenario (same options + inactive icon - orange dot)

- If the user sets a new recurring amount, we revoke the previous grant and create a new one with the specified amount
- If the user wants to set a one time amount, we need to keep the previous grant, create a new one for the specified amount and use the one time grant for all future payments until the recurring amount is available again
- When the recurring amount is available again we need to sum up the remaining amounts and we use recurring one until it runs out of funds again - no need to show the out_of_funds screen if we still have funds in the one time amount grant.

TODO:
- [x] https://github.com/interledger/web-monetization-extension/issues/376
- [x] https://github.com/interledger/web-monetization-extension/issues/425
- [x] https://github.com/interledger/web-monetization-extension/issues/377

Out of scope for [beta](https://github.com/interledger/web-monetization-extension/milestone/2) release:
- https://github.com/interledger/web-monetization-extension/issues/444
- https://github.com/interledger/web-monetization-extension/issues/445
- https://github.com/interledger/web-monetization-extension/issues/446

sidvishnoi:
Notes: 
- https://interledgerfoundation.slack.com/archives/C054B631B97/p1718799633709149

tselit:
Blocked by #377  and the OP depedency (@tselit to update with outlook for designs after chatting to Madalina)

PR #379, opened 2024-06-28
feat!: store both recurring/one-time grants & spents; add `state`

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/376

1. Update storage as suggested in https://github.com/interledger/web-monetization-extension/issues/376, but with following differences:
   - Instead of storing `interval` and `approvedAmount` as separate keys in grants, store them as `amount: WalletAmount` (but within the grant)
   - Instead of storing balances, we'll store spent amounts as API only returns spent amounts.
   - Add `GrantDetails.type = "recurring" | "one-time"` to make it easier for TypeScript to discriminate between two.
2. Store a `Storage.version` to help with migrations, keep migrations indexed by this sequential version number.
3. While we're doing a migration here:
   - Add `Storage.state`. This is also a precursor to https://github.com/interledger/web-monetization-extension/issues/144 (`out_of_funds` state)
   - Replace `Storage.hasHostPermissions` with `Storage.state = "missing_host_permissions"` 

Also includes some parts of https://github.com/interledger/web-monetization-extension/issues/144 (setting the first available grant, but without `switchGrant`) and https://github.com/interledger/web-monetization-extension/issues/272.


raducristianpopa:

<table role="table">
<thead>
  <tr>
    <th>Name</th>
    <th>Link</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td align="center">0514f88aa10a0e2ba358734b9bf092812d825e52</td>
  </tr>
  <tr>
    <td align="center"><a href="https://github.com/interledger/web-monetization-extension/actions/runs/9711016075">Run #9711016075</a></td>
  </tr>
  <tr><td align="center">Badge</td><td align="center">N/A</td></tr><tr><td align="center">Badge</td><td align="center">N/A</td></tr>
</tbody>
</table>

sidvishnoi:
Before the `Storage.state` commit, changed storage like this, seamlessly on extension update:

Screenshot from 2024-06-28 14-33-34

src/background/services/openPayments.ts
```diff
@@ -330,25 +338,41 @@ export class OpenPaymentsService {
...
+        recurringGrant: grantDetails,
+        recurringGrantRemainingBalance: amount
+      })
+    } else {
+      await this.storage.set({
+        ...data,
+        oneTimeGrant: grantDetails,
+        oneTimeGrantRemainingBalance: amount
+      })
+    }
```
sidvishnoi, line 373: Nit: Had to be verbose here as TypeScript was causing issues with 
```ts
const data = {
  // ... common properties
  ...(type === "recurring" ? {recurringGrant: grantDetails} : {oneTimeGrant: grantDetails})
```

src/background/services/openPayments.ts
```diff
@@ -330,25 +338,41 @@ export class OpenPaymentsService {
...
+    this.grant = grantDetails
+    this.token = this.grant.accessToken
```
sidvishnoi, line 375: Nit: Keeping `this.token` to avoid unrelated changes (we can use this.grant.accessToken` everywhere now).

src/background/services/storage.ts
```diff
@@ -1,16 +1,23 @@
...
+  GrantDetails,
+  Storage,
+  StorageKey,
+  WalletAmount
+} from '@/shared/types'
 import { type Browser } from 'webextension-polyfill'
 import { EventsService } from './events'
 
 const defaultStorage = {
+  version: 2,
```
sidvishnoi, line 11: For migrations, change this version and add a migration script in `MIGRATIONS` below.
New additions that can be dynamically set don't need migrations (e.g. we check if value is null etc.) but other structural changes would need migrations for keeping compatibility with existing installations. Also see the comment in `MIGRATIONS`.

sidvishnoi: I'll add this comment in code.

src/shared/types.ts
```diff
@@ -13,41 +13,60 @@ export interface Amount {
...
+  amount: Omit<WalletAmount, 'interval'>
+}
+export interface RecurringGrant extends GrantDetailsBase {
+  type: 'recurring'
+  amount: Required<WalletAmount>
 }
+export type GrantDetails = OneTimeGrant | RecurringGrant
+
+/** Bigint amount, before transformation with assetScale */
+type AmountValue = string
```
raducristianpopa, line 37: Let's use this for `WalletAmount['value']` and `Amount['value']` as well.

package.json
```diff
@@ -104,7 +104,7 @@
     "pnpm": "^9.4.0",
     "npm": "pnpm",
     "yarn": "pnpm",
-    "node": "20.14.0"
+    "node": "^20.14"
```
raducristianpopa, line 107: See my comment in #380.

src/background/services/background.ts
```diff
@@ -154,28 +155,37 @@ export class Background {
...
     })
   }
 
   checkPermissions = async () => {
     try {
       this.logger.debug('checking hosts permission')
-      const status = await this.browser.permissions.contains(PERMISSION_HOSTS)
-      this.storage.setHostPermissionStatus(status)
+      const ok = await this.browser.permissions.contains(PERMISSION_HOSTS)
+      this.storage.setState({ missing_host_permissions: !ok })
```
raducristianpopa, line 188:
```suggestion
      const hasPermissions = await this.browser.permissions.contains(PERMISSION_HOSTS)
      this.storage.setState({ missing_host_permissions: !hasPermissions })
```

src/background/services/events.ts
```diff
@@ -1,8 +1,12 @@
 import { EventEmitter } from 'events'
+import type { Storage } from '@/shared/types'
 
 interface BackgroundEvents {
   'storage.rate_of_pay_update': { rate: string }
-  'storage.host_permissions_update': { status: boolean }
+  'storage.state_update': {
+    state: Storage['state']
+    prevState: Storage['state']
+  }
```
raducristianpopa, line 9: Why are we passing the current state and the previous state?

src/background/services/background.ts
```diff
@@ -154,28 +155,37 @@ export class Background {
...
+      const existingData = await this.storage.get()
+      this.logger.info(existingData)
       if (details.reason === 'install') {
         await this.storage.populate()
         await this.openPaymentsService.generateKeys()
       }
+      const migrated = await this.storage.migrate()
+      if (migrated) {
+        const prevVersion = existingData.version ?? 1
+        this.logger.info(`Migrated from ${prevVersion} to ${migrated.version}`)
```
raducristianpopa, line 178: I know we are explicitly checking for the version in the `migrate` method to prevent migrating multiple times, but let's check for `details.reason === 'update'` as well.

src/background/services/background.ts
```diff
@@ -154,28 +155,37 @@ export class Background {
...
-      this.logger.info(await this.storage.get())
+      const existingData = await this.storage.get()
+      this.logger.info(existingData)
       if (details.reason === 'install') {
         await this.storage.populate()
         await this.openPaymentsService.generateKeys()
       }
+      const migrated = await this.storage.migrate()
+      if (migrated) {
+        const prevVersion = existingData.version ?? 1
```
raducristianpopa, line 177: Opinions on setting the storage version with Webpack defines?

src/background/services/background.ts
```diff
@@ -154,28 +155,37 @@ export class Background {
...
+    this.events.on('storage.state_update', async ({ state, prevState }) => {
+      this.logger.info('state changed', { state, prevState })
       // TODO: change icon here in future
     })
   }
 
   bindOnInstalled() {
     this.browser.runtime.onInstalled.addListener(async (details) => {
-      this.logger.info(await this.storage.get())
+      const existingData = await this.storage.get()
```
raducristianpopa, line 169:
```suggestion
      const data = await this.storage.get()
```

src/background/services/openPayments.ts
```diff
@@ -459,18 +483,17 @@ export class OpenPaymentsService {
   }
 
   async disconnectWallet() {
-    const { grant } = await this.storage.get(['grant'])
+    const { recurringGrant, oneTimeGrant } = await this.storage.get([
+      'recurringGrant',
+      'oneTimeGrant'
+    ])
+    const grant = recurringGrant || oneTimeGrant
```
raducristianpopa, line 490: We should try to revoke both grants of wallet disconnect. For example:
- User sets a recurring amount when connecting the wallet
- Runs out of funds and adds a one-time budget
- User disconnects the wallet.

In this case, we will only revoke the recurring grant, but not the one-time one.

src/background/services/background.ts
sidvishnoi: It's in defaultStorage as we'll update version typically when we update values in defaultStorage (or its structure). Having it in defines won't make sense here, as it's a fixed part of code.

src/background/services/openPayments.ts
sidvishnoi: This is something I'm leaving for the other PR, when we'd have logic like switchGrant and "top up" one time over existing grant. As of this PR, we'd only have one active grant at a time.

src/background/services/background.ts
sidvishnoi: Was just avoiding adding newlines 😁

src/background/services/events.ts
sidvishnoi: Helpful with logging/metrics to know if users get into multiple errors states one after other. Also, we can add transitions between states in popup/icons.

src/background/services/openPayments.ts
raducristianpopa: > This is something I'm leaving for the other PR, when we'd have logic like switchGrant and "top up" one time over existing grant. As of this PR, we'd only have one active grant at a time.

Add a comment for it.

src/background/services/storage.ts
```diff
@@ -1,16 +1,30 @@
...
+  StorageKey,
+  WalletAmount
+} from '@/shared/types'
 import { type Browser } from 'webextension-polyfill'
 import { EventsService } from './events'
 
 const defaultStorage = {
+  /**
+   * For migrations, increase this version and add a migration script in
+   * {@linkcode MIGRATIONS} below. New additions to structure that can be
```
sidvishnoi, line 13: Aside: In VSCode, MIGRATIONS is linked (can be followed on click)
image

dianafulga (approved):

sidvishnoi:
@raducristianpopa @dianafulga In open outgoingPayment, we actually get `grantSpentDebitAmount` instead of balance. Should we compute balance on each outgoing payment, or just store spentAmount instead and compute as needed?

src/background/services/storage.ts
```diff
@@ -31,10 +32,25 @@ const defaultStorage = {
...
+
   constructor(
     private browser: Browser,
     private events: EventsService
-  ) {}
+  ) {
+    const bigIntMin = (a: string, b: string) => (BigInt(a) < BigInt(b) ? a : b)
+    this.setRemainingBalanceRecurring = new DebounceWithQueue(
+      (amount: string) => this.set({ recurringGrantRemainingBalance: amount }),
+      (args) => [args.reduce((m, [e]) => (!m ? e : bigIntMin(m, e)), '')],
```
sidvishnoi, line 45: If we're getting multiple calls to `Storage.setRemainingBalance` (which maybe out of order depending on which createOutgoingPayment resolves first), we can collect the calls and update storage less often.
It might be ok to update balance in storage on each createOutgoingPayment also - maybe just unnecessary optimization, but I think more meaningful when we have like 5-6+ wallet addresses per page.

dianafulga:
> @raducristianpopa @dianafulga In open outgoingPayment, we actually get `grantSpentDebitAmount` instead of balance. Should we compute balance on each outgoing payment, or just store spentAmount instead and compute as needed?

From what I know, we shouldn't compute anything on our side, it Should come from the API. Wait until it's implemented.

sidvishnoi:
It appears it was decided in https://github.com/interledger/open-payments/issues/469 that they'd return spent amount only instead of balance (see edit in issue description).

So, now it's upto us how to use it. We already have what amount is approved in grant. Now, we can either subtract it on each storage, or only when we need the balance. I think if we don't need to know balance to know if we can make a payment (like before each payment or after, to reach the `out_of_founds` state), we can store spentAmount and compute balance as needed (for UI only I believe).

src/shared/helpers.ts
```diff
@@ -88,6 +88,48 @@ export function debounceAsync<T extends unknown[], R extends Promise<unknown>>(
...
+ *   (total: number) => saveToStorage(total),
+ *   (collectedArgs) => collectedArgs.reduce(total, [val] => total + val, 0),
+ *   wait
+ * )
+ * debounceWithQueue.enqueue(10)
+ * debounceWithQueue.enqueue(15)
+ * // results in saveToStorage(25)
+ * ```
+ */
+export class DebounceWithQueue<Args extends unknown[]> {
```
raducristianpopa, line 108: What happens if the user closes the browser and we still have items in the queue?

sidvishnoi: They may get lost (upto a second of queue), but we can get latest data on next payment.

Anyway, I was testing this out and will move it into a separate PR when we store spent/balances on payment.

raducristianpopa (approved):

Issue #425, opened 2024-07-15
[TASK] Ability to switch grants (when recurring fails -> one time -> out_of_funds)


sidvishnoi:
Part of https://github.com/interledger/web-monetization-extension/issues/144

Issue #377, opened 2024-06-27
Add UI for "out of funds" error state


sidvishnoi:
It would have:
- an error message
- an option to specify amount + recurring flag

tselit:
The UI designs are ready for review by Madalina, before we continue with Development:
[link to "out of funds" designs on Figma](https://www.figma.com/design/0S8Oj8ZB5MvkGtIrVRswDH/Web-monetization?node-id=509-852&m=dev)

sidvishnoi:
From [Slack](https://interledgerfoundation.slack.com/archives/C06CVJRT5N3/p1720092323159379):

> An important question is when to show the out of funds error, is it when they are about to make a payment or if they open the pop up or both - @RabebOthmani 

@sidvishnoi: I think there's two parts here: telling user they're out of funds and second is their action. The error part should be shown only when we can do nothing more without user interaction. We can show the action screen in various scenarios:
- Can do nothing without adding funds (force show this screen)
- One-time pay (should warn user, instead of forcing to add funds)
   - Can make smaller continuous payments, but can't make a larger single payment (not enough funds, but not out of funds)
   - Could make a partial payment (to multiple addresses), just not the total amount user asked for (not enough funds again)
- I think they should be able to manually take action even before running out of funds, that's something we haven't discussed before, so can ignore for now

tselit:
fyi @RabebOthmani and @sidvishnoi 
Yesterday (18 July), Madalina reviewed & was happy with these proposed screens, from the design & identity point of view.

PR #426, opened 2024-07-15
feat: allow adding funds when out of funds; add `switchGrant`

sidvishnoi:
## Context

- Closes https://github.com/interledger/web-monetization-extension/issues/425
- Part of https://github.com/interledger/web-monetization-extension/issues/144
- Part of https://github.com/interledger/web-monetization-extension/issues/147

1. Add backend for supporting both recurring and one-time grants in parallel.
   1. Only one grant is used at a time.
   1. Recurring grant is preferred over one-time grant.
   1. Add `addFunds` function, reusing much of the existing `connectWallet` logic (extracted `completeGrant` which was common to both).
   1. We cancel existing grants as described in https://github.com/interledger/web-monetization-extension/issues/144'
   1. We can switch which one to use with `OpenPayments.switchGrant()`
      1. A payment session can call `switchGrant` when a payment fails with "out of balance" error.
      1. This function has no parameters. OpenPayments service decides which grant to use next.
      1. We maintain an internal-state `isGrantUsable.{oneTime,recurring}` to know which grant has been tried already or is available to use.
1. Add `out_of_funds` state
   1. If payments fail with "out of balance" error, and if no grant is usable anymore we move to this state.
   1. Added basic UI to test adding grant works. Will do real UI as a follow-up PR (https://github.com/interledger/web-monetization-extension/issues/377)
   1. We reset this state on extension start, so we can try payments to know current balance status. We also reset it when recurringGrant renews (via alarms API).
1. Smaller changes:
   1. Made `OpenPayments.grant` a getter/setter, so we can add debuglogging. `OpenPayments.token` is synced with this grant now.
   1. Given `Storage.setState({ out_of_funds: false })` is called on each successful payment (how else we know if we can make a payment now, without actually making it? calculating balance is bit fragile), adding a little caching layer in `Storage.setState`.


raducristianpopa:

<table role="table">
<thead>
  <tr>
    <th>Name</th>
    <th>Link</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td align="center">0788277de2c68201c131f26baae55cbbbc8577c0</td>
  </tr>
  <tr>
    <td align="center"><a href="https://github.com/interledger/web-monetization-extension/actions/runs/9938419678">Run #9938419678</a></td>
  </tr>
  <tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/26009108449/artifacts/1701457585">Download</a></td></tr><tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/26009108449/artifacts/1701457590">Download</a></td></tr>
</tbody>
</table>

sidvishnoi:
A demo:

[Screencast from 18-07-24 03:47:50 PM IST.webm](https://github.com/user-attachments/assets/65b4906a-ab74-43a9-9a03-5aa851c7d7e8)

src/background/services/background.ts
```diff
@@ -31,6 +31,9 @@ export class Background {
   ) {}
 
   async start() {
+    // Reset out_of_funds state, we'll detect latest state as we make a payment.
+    await this.storage.setState({ out_of_funds: false })
+
```
sidvishnoi, line 36: This will get called only on background start. Fine if background idles after some time, but not if it stays alive. Need a way to get out of out_of_funds when recurring balance renews!

sidvishnoi: If continuous payments are enabled, we can make a payment and know we have funds now. But not without it.

sidvishnoi: Also, TODO: reset the `out_of_funds` on first payment if it was already not reset (don't want to call it on each payment).

src/background/services/storage.ts
```diff
@@ -128,12 +131,13 @@ export class StorageService {
...
   async setState(state: Storage['state']): Promise<boolean> {
-    const { state: prevState } = await this.get(['state'])
+    const prevState = this.currentState ?? (await this.get(['state'])).state
 
-    const newState: Storage['state'] = { ...prevState }
+    const newState: Storage['state'] = { ...this.currentState }
     for (const key of Object.keys(state) as ExtensionState[]) {
       newState[key] = state[key]
     }
+    this.currentState = newState
```
sidvishnoi, line 140: Happy with this bit of caching?

src/popup/components/OutOfFunds.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react'
```
sidvishnoi, line 1: This will be reworked in https://github.com/interledger/web-monetization-extension/issues/377 next.

sidvishnoi: But I've added the boilerplate and integration.

src/background/services/monetization.ts
```diff
@@ -317,9 +353,13 @@ export class MonetizationService {
...
 
     return {
-      ...storedData,
+      ...dataFromStorage,
       balance: balance.total.toString(),
       url,
+      grants: {
+        oneTime: oneTimeGrant?.amount,
+        recurring: recurringGrant?.amount
+      },
```
raducristianpopa, line 362: Why do we need to send the amounts as well? Aren't they computed within `this.storage.getBalance` ?

src/background/services/openPayments.ts
```diff
@@ -92,16 +97,37 @@ type TabUpdateCallback = Parameters<Tabs.onUpdatedEvent['addListener']>[0]
 export class OpenPaymentsService {
   client?: AuthenticatedClient
 
+  public switchGrant: OpenPaymentsService['_switchGrant']
+
   private token: AccessToken
-  private grant: GrantDetails | null
+  /** Whether a grant has enough balance to make payments */
+  private isGrantUsable = { recurring: false, oneTime: false }
```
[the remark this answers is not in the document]
sidvishnoi: This rename will affect a lot of places 😄

src/background/services/openPayments.ts
```diff
@@ -360,6 +388,22 @@ export class OpenPaymentsService {
...
+        oneTimeGrant: null,
+        oneTimeGrantSpentAmount: '0'
+      })
+    } else if (grants.recurringGrant && grantDetails.type === 'recurring') {
+      await this.cancelGrant(grants.recurringGrant.continue)
+      await this.storage.set({
+        recurringGrant: null,
+        recurringGrantSpentAmount: '0'
+      })
+    }
```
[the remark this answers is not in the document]
sidvishnoi: Hmm. I'll extract this into a function and reuse it in connectWallet, as connectWallet is mostly same except those checks. Though, I think initializing client again is fine fine?

src/background/services/openPayments.ts
```diff
@@ -582,6 +641,40 @@ export class OpenPaymentsService {
...
+      }
+    } else if (this.grant?.type === 'one-time') {
+      this.isGrantUsable.oneTime = false
+      if (
+        recurringGrant &&
+        getNextOccurrence(recurringGrant.amount.interval).valueOf() <=
+          Date.now()
+      ) {
+        this.grant = recurringGrant
+        return 'recurring'
```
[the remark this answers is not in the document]
sidvishnoi: It's a sanity check. If we are using one time grant, and we run out of funds, we at a last resort try to use recurringGrant if possible.
The test is specifically checking if renewal.... hmmm the test seems wrong.. lol.

src/background/services/background.ts
```diff
@@ -40,6 +49,25 @@ export class Background {
     this.sendToPopup.start()
   }
 
+  async scheduleResetOutOfFundsState() {
+    // Reset out_of_funds state, we'll detect latest state as we make a payment.
+    await this.storage.setState({ out_of_funds: false })
```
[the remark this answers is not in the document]
raducristianpopa: Oh, the alarm will be deleted since it's session based. That's why we are resetting the state.

src/background/services/monetization.ts
sidvishnoi: I'm passing the amount to UI so we can show a default value in input where user will add funds, and make it possible also to suggest amounts.
The recurring one also includes interval, which we use to show when will the money be added automatically for use, if the user chooses to do nothing.
I'm passing raw inputs so we can compute those details as and when needed in the popup. I'm passing just the amounts as that's what we need in popup, and avoiding passing tokens to reduce security risks.

src/background/services/background.ts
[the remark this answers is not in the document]
sidvishnoi: I'll check when the alarm really goes away - is it background script's kill or browser session kill.

src/background/services/openPayments.ts
```diff
@@ -92,16 +97,37 @@ type TabUpdateCallback = Parameters<Tabs.onUpdatedEvent['addListener']>[0]
...
 
   constructor(
     private browser: Browser,
     private storage: StorageService,
     private deduplicator: Deduplicator,
+    private logger: Logger,
     private t: Translation
   ) {
     void this.initialize()
+    this.switchGrant = this.deduplicator.dedupe(this._switchGrant.bind(this))
```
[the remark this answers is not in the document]
sidvishnoi: Do you mean like following?
```ts
public switchGrant = this.deduplicator.dedupe(async function() {
  // ...
})
```

src/popup/components/OutOfFunds.tsx
sidvishnoi: Note that solution for #377 will be added after this PR is merged. But feel free to ignore the UI part in this PR.

src/background/services/openPayments.ts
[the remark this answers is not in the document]
raducristianpopa: Yes, or:
```ts
public switchGrant(...args) {
  return this.deduplicator.dedupe(...)
}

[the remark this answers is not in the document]
Comment (sidvishnoi) [reply src/background/services/openPayments.ts:114]: Just to clarify, the `_switchGrant` will stay, right? Because otherwise on each call we'll have different instance of callback passed to dedupe.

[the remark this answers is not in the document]
Comment (sidvishnoi) [reply src/background/services/openPayments.ts:114]: Note that we can call dedupe from PaymentSession, but then that becomes responsibility of caller, which is something I try to avoid.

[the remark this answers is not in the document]
Comment (raducristianpopa) [reply src/background/services/openPayments.ts:114]: Oh, you're right. Let's keep this as is.

File: src/background/services/openPayments.ts (hunk 6 above)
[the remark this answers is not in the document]
Comment (sidvishnoi) [reply src/background/services/openPayments.ts]: Called in `addFunds` instead of topUp.

File: src/background/services/openPayments.ts [hunk 10]
```diff
@@ -326,13 +326,68 @@ export class OpenPaymentsService {
...
+        oneTimeGrant: null,
+        oneTimeGrantSpentAmount: '0'
+      })
+    } else if (grants.recurringGrant && recurring) {
+      await this.cancelGrant(grants.recurringGrant.continue)
+      await this.storage.set({
+        recurringGrant: null,
+        recurringGrantSpentAmount: '0'
+      })
+    }
```
raducristianpopa, line 358: Let's move this after we call `completeGrant`. In case something fails we are not going to clear the existing grants.

sidvishnoi: `completeGrant` sets these in storage, hmm. Yeah we just need to cancel grants then.

src/background/services/openPayments.ts
```diff
@@ -298,13 +326,60 @@ export class OpenPaymentsService {
...
+      recurring
+    )
+
+    // cancel existing grants of same type, if any
+    const grants = await this.storage.get(['oneTimeGrant', 'recurringGrant'])
+    if (grants.oneTimeGrant && !recurring) {
+      await this.cancelGrant(grants.oneTimeGrant.continue)
+    } else if (grants.recurringGrant && recurring) {
+      await this.cancelGrant(grants.recurringGrant.continue)
+    }
```
raducristianpopa, line 356: Isn't this going to cancel the new grant? If `completeGrant` already updated the storage with the new one, we are going to retrieve the grant from storage and cancel them immediately.

I feel like the logic should be:
1. retrieve current grant from storage based on `recurring`;
2. call `completeGrant` - which already performs `.set()` on the storage;
3.  cancel previous grant (if it exists) - we already have the details since we retrieved them at step 1

src/background/services/background.ts
```diff
@@ -90,6 +118,10 @@ export class Background {
 
             case PopupToBackgroundAction.CONNECT_WALLET:
               await this.openPaymentsService.connectWallet(message.payload)
+              await this.browser.alarms.clear(ALARM_RESET_OUT_OF_FUNDS)
```
raducristianpopa, line 121: Since `CONNECT_WALLET` is only called during setup, there is no point in clearing the alarm here. We can clear the alarm on `DISCONNECT_WALLET`.

src/background/services/openPayments.ts
```diff
@@ -92,16 +97,37 @@ type TabUpdateCallback = Parameters<Tabs.onUpdatedEvent['addListener']>[0]
...
+
+  private get grant() {
+    return this.grantDetails
+  }
+  private set grant(grantDetails) {
+    this.logger.debug(`🤝🏻 Using grant: ${grantDetails?.type || null}`)
+    this.grantDetails = grantDetails
+    this.token = grantDetails
+      ? grantDetails.accessToken
+      : { value: '', manageUrl: '' }
```
raducristianpopa, line 130:
```suggestion
  private get grant() {
    return this.grantDetails
  }
  
  private set grant(grantDetails) {
    this.logger.debug(`🤝🏻 Using grant: ${grantDetails?.type || null}`)
    this.grantDetails = grantDetails
    this.token = grantDetails
      ? grantDetails.accessToken
      : { value: '', manageUrl: '' }
```

src/background/services/background.ts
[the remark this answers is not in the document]
sidvishnoi: From my debugging, the alarm never went away, even on browser close! 😆 

Let's keep this as is, as it's not guaranteed when alarm is removed.
> Alarms generally persist until an extension is updated. However, this is not guaranteed, and alarms **may** be cleared when the browser is restarted (emphasis mine)

src/background/services/openPayments.ts
```diff
@@ -298,13 +326,60 @@ export class OpenPaymentsService {
...
+      await this.cancelGrant(grants.oneTimeGrant.continue)
+    } else if (grants.recurringGrant && recurring) {
+      await this.cancelGrant(grants.recurringGrant.continue)
+    }
+
+    // If we already have a recurring grant, and user adds a one-time popup,
+    // prefer existing recurring grant. If the recurring grant doesn't have
+    // enough funds, which is likely the reason to add a one-time popup (but not
+    // always, as user can add top-up before running out of funds in future),
+    // we'll switch to the one-time grant when we try to make a payment.
```
raducristianpopa, line 362: I understand your concern for the future, but right now it would make the most sense to automatically switch to `one-time` if the user already has a recurring grant and ran out of funds.

src/background/services/openPayments.ts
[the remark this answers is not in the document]
raducristianpopa: Let me know if my understanding is right.

Context:
- User has a recurring grant with available funds
- User has a leftover one-time grant with available funds

Scenario:
- User keeps visiting WM websites and runs out of funds for the recurring one
- We switch to the one time one
- If the user runs out of funds again (for the one time one), we are trying to switch back to the recurring one

Wouldn't this lead to an infinite loop? `recurring` -> `oneTime` -> `recurring` -> `oneTime`

src/background/services/paymentSession.ts
```diff
@@ -161,15 +165,20 @@ export class PaymentSession {
         } else if (isTokenExpiredError(e)) {
           await this.openPaymentsService.rotateToken()
           continue
+        } else if (isOutOfBalanceError(e)) {
+          const switched = await this.openPaymentsService.switchGrant()
+          if (switched === null) {
+            this.events.emit('open_payments.out_of_funds')
+          } else {
+            await this.setIncomingPaymentUrl()
+          }
```
raducristianpopa, line 174: Why are we creating a new incoming payment?

src/background/services/openPayments.ts
sidvishnoi: Oops! great catch!

src/background/services/paymentSession.ts
raducristianpopa: Better context with lines:
https://github.com/interledger/web-monetization-extension/blob/7e318b8937dc3065707c42f7e2e33ac796ff88a3/src/background/services/paymentSession.ts#L168-L175

src/background/services/openPayments.ts
sidvishnoi: What's changed here exactly? added a new line after getter?

raducristianpopa: Yep.

src/background/services/openPayments.ts
sidvishnoi: Switch to whatever the new grant is, got it.

src/background/services/paymentSession.ts
sidvishnoi: Oh, I saw accessToken used there and thought it was the one from grant we created. Will fix.

src/background/services/openPayments.ts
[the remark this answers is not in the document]
sidvishnoi: Not infinite loop as we have `this.isGrantUsable`.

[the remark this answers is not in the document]
raducristianpopa: Oh, okay. Totally missed that.

Still confused by that check:
```ts
      if (
        recurringGrant &&
        getNextOccurrence(recurringGrant.amount.interval).valueOf() <=
          Date.now()
      )
```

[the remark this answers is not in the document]
sidvishnoi: We're trying to switch back to recurring one only if grant is available now (logic to be fixed do detect is grant is renewed).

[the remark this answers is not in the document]
sidvishnoi: Yeah the logic is wrong. In fact, it's not even needed.

`getNextOccurrence` here was returning next occurrence date after "now", and then comparing it to "now" again. The goal was to find if we've passed the date from when the we ran out of funds.

Will remove the check, as we can know if out of funds on next payment either way.

dianafulga (approved):

raducristianpopa (approved):

Issue #375, opened 2024-06-27
[BUG] WM rate of pay not consistent


sidvishnoi:
Discussed with @raducristianpopa yesterday.

When sending quicker payments, we await the createPayment call, which takes its own time to process. Then we wait for "interval" (without subtracting the time it took to process the payment). So, payments are slower than user configured.

sidvishnoi:
[Rafiki v1.0.0-alpha.16](https://github.com/interledger/rafiki/releases/tag/v1.0.0-alpha.16) improved performance a lot. But there's still some slowdown when sending multiple concurrent requests. Looking forward to more improvements before we move to this.

PR #435, opened 2024-07-17
refactor(background/PaymentSession): extract `payContinuous`

dianafulga:
Part of #375 

---

If closing issue #375:

We ignore if the payment is successful and wait anyway the right amount of time for the next payment, because otherwise we need to await the response for the payment which is the problem reported here.
Computing the difference it's a rather shady implementation, because if the interval is 1s and the api call takes 4s, then there is nothing we can do.


raducristianpopa:

<table role="table">
<thead>
  <tr>
    <th>Name</th>
    <th>Link</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td align="center">8f6c3e6dee2a28919abceadcd2d22e7e24146aa3</td>
  </tr>
  <tr>
    <td align="center"><a href="https://github.com/interledger/web-monetization-extension/actions/runs/9981555154">Run #9981555154</a></td>
  </tr>
  <tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/26127895859/artifacts/1712620642">Download</a></td></tr><tr><td align="center">Badge</td><td align="center"><a href="https://github.com/interledger/web-monetization-extension/suites/26127895859/artifacts/1712620644">Download</a></td></tr>
</tbody>
</table>

src/background/services/paymentSession.ts
```diff
@@ -149,35 +147,15 @@ export class PaymentSession {
     await sleep(waitTime)
 
     while (this.active) {
-      try {
-        outgoingPayment = await this.openPaymentsService.createOutgoingPayment({
+      this.openPaymentsService
```
sidvishnoi, line 150: How about we move this into a private function and avoid the `.then()`? Will be clearer IMO:
```ts
while (this.active) {
  void this.foo(...)
  await sleep(...)
}
```

src/background/services/paymentSession.ts
```diff
@@ -336,4 +274,57 @@ export class PaymentSession {
...
+      } else if (e instanceof OpenPaymentsClientError) {
+        // We need better error handling.
+        if (e.status === 400) {
+          await this.setIncomingPaymentUrl()
+        }
+
+        // TODO: Check what Rafiki returns when there is no amount
+        // left in the grant.
+        throw new Error(e.message)
+      }
```
sidvishnoi, line 327: In case of errors, I wonder if there should be a way to pause other payments momentarily, so they don't trigger same errors over and over until they're handled? What do you think @raducristianpopa? Or is that fine as is?

src/background/services/paymentSession.ts
```diff
@@ -336,4 +274,57 @@ export class PaymentSession {
...
+      })
+
+      // TO DO: find a better source of truth for deciding if overpaying is applicable
+      if (this.intervalInMs > 1000) {
+        this.tabState.saveOverpaying(this.tab, this.url, {
+          walletAddressId: this.receiver.id,
+          monetizationEvent: monetizationEventDetails,
+          intervalInMs: this.intervalInMs
+        })
+      }
```
sidvishnoi, line 312: Should we move these out of `try catch {}`? Just to be clear that these can't throw?

dianafulga: for me this doesn't particulary make sense to move them out, since they are dependent on the outgoingPayment being received successfully

src/background/services/paymentSession.ts
dianafulga: pausing them would reintroduce the difference of frequency

src/background/services/paymentSession.ts
sidvishnoi: That I understand, but just meant as code style.
```ts
const outgoingPayment = await create(...).catch(error => {...})

if (!outgoingPayment) return

sendMonetizationEvent()
...
```

src/background/services/paymentSession.ts
raducristianpopa: I'm fine with the implementation as is for now. We can revisit this in the future.

sidvishnoi:
Sorry Diana for this merge conflict 😬

src/background/services/paymentSession.ts
```diff
@@ -343,4 +278,57 @@ export class PaymentSession {
...
+    } catch (e) {
+      if (isKeyRevokedError(e)) {
+        this.events.emit('open_payments.key_revoked')
+      } else if (isTokenExpiredError(e)) {
+        this.openPaymentsService.rotateToken()
```
sidvishnoi, line 321:
```suggestion
        await this.openPaymentsService.rotateToken()
```

src/background/services/paymentSession.ts
```diff
@@ -343,4 +278,57 @@ export class PaymentSession {
...
+      } else if (e instanceof OpenPaymentsClientError) {
+        // We need better error handling.
+        if (e.status === 400) {
+          await this.setIncomingPaymentUrl()
```
sidvishnoi, line 325: Should dedupe this call I think.

src/background/services/paymentSession.ts
```diff
@@ -343,4 +278,57 @@ export class PaymentSession {
...
+        }
+
+        // TODO: Check what Rafiki returns when there is no amount
+        // left in the grant.
```
sidvishnoi, line 329: Handled already
```suggestion
```

src/background/services/paymentSession.ts
```diff
@@ -343,4 +278,57 @@ export class PaymentSession {
...
+          monetizationEvent: monetizationEventDetails,
+          intervalInMs: this.intervalInMs
+        })
+      }
+    } catch (e) {
+      if (isKeyRevokedError(e)) {
+        this.events.emit('open_payments.key_revoked')
+      } else if (isTokenExpiredError(e)) {
+        this.openPaymentsService.rotateToken()
+      } else if (e instanceof OpenPaymentsClientError) {
```
sidvishnoi, line 322: Lost in merge...
```suggestion
      } else if (isOutOfBalanceError(e)) {
          const switched = await this.openPaymentsService.switchGrant()
          if (switched === null) {
            this.events.emit('open_payments.out_of_funds')
          }
      } else if (e instanceof OpenPaymentsClientError) {
```

src/background/services/paymentSession.ts
raducristianpopa: I don't think it needs to be deduped, but we might have a small bug. If the `incomingPaymentUrl` is set, the `setIncomingPaymentUrl` is going to return only. If the incoming payment expires, a new one will not be created.

raducristianpopa:
A lot of outgoing payments are failing for me when sending a payment every second (asset scale 9 to asset scale 9).

Can you confirm the behaviour @dianafulga @sidvishnoi?

raducristianpopa:
> A lot of outgoing payments are failing for me when sending a payment every second (asset scale 9 to asset scale 9).
> 
> Can you confirm the behaviour @dianafulga @sidvishnoi?

@dianafulga

dianafulga:
> > A lot of outgoing payments are failing for me when sending a payment every second (asset scale 9 to asset scale 9).
> > Can you confirm the behaviour @dianafulga @sidvishnoi?
> 
> @dianafulga

Yes, most of the queries are getting canceled.

raducristianpopa:
Let's postpone this for now and move it to out of beta.

sidvishnoi:
Decided to keep this change as a refactor so we don't have to keep updating this PR. We'll keep the "awaiting" the payments for now, and when we're ready to send on shorter interval, we'll just stop awaiting.

src/background/services/paymentSession.ts
```diff
@@ -186,71 +184,12 @@ export class PaymentSession {
...
-              currency: receiveAmount.assetCode,
-              value: transformBalance(
-                receiveAmount.value,
-                receiveAmount.assetScale
-              )
-            },
-            incomingPayment,
-            paymentPointer: this.receiver.id
-          }
+    await sleep(waitTime)
```
raducristianpopa, line 187: Can be removed since `sleep` was moved in the `if` block from above.
```suggestion
```

sidvishnoi (approved):

sidvishnoi (approved):

sidvishnoi (changes requested):
Wait...

sidvishnoi (approved):
Ok, all good. was concerned about `this.waiting`.

Issue #499, opened 2024-08-12
Move MessageManager in the container


raducristianpopa:
### Context

Right now we are defining the messages outside of the container. We would like them to be registered in the container as well.

### Todos

- [ ] Register the message manager in the container

PR #534, opened 2024-08-22
refactor: use `message.send()` directly, move message to container

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/499

- Move `MessageManager` instance into container/context so we don't have to import `browser` at top-level (makes browser easily replaceable for testing)
- Reduce indirection when sending messages, and simplify call signature - aligning it more with how we use `SendToPopup`


src/content/services/monetizationTagManager.ts
```diff
@@ -94,7 +89,7 @@ export class MonetizationTagManager extends EventEmitter {
   }
 
   private async resumeAllMonetization() {
-    const response = await isWMEnabled()
+    const response = await this.message.send('IS_WM_ENABLED', undefined)
```
raducristianpopa, line 92: Look if we can avoid passing undefined as the second parameter (the payload).

src/popup/components/ErrorKeyRevoked.tsx
```diff
@@ -11,8 +11,8 @@ import type { Response } from '@/shared/messages'
 
 interface Props {
   info: Pick<PopupStore, 'publicKey' | 'walletAddress'>
-  disconnectWallet: () => Promise<Response>
-  reconnectWallet: () => Promise<Response>
+  disconnectWallet: () => Promise<Response<void>>
```
raducristianpopa, line 14: Check if we can avoid passing void to the `Response` type.

src/popup/components/OutOfFunds.tsx
```diff
@@ -68,7 +68,7 @@ interface AddFundsProps {
   info: Pick<WalletAddress, 'id' | 'assetCode' | 'assetScale'>
   recurring: boolean
   defaultAmount: AmountValue
-  requestAddFunds: (details: AddFundsPayload) => Promise<Response>
+  requestAddFunds: (details: AddFundsPayload) => Promise<Response<void>>
```
raducristianpopa, line 71: Ditto above.

src/content/services/monetizationTagManager.ts
sidvishnoi: Done in https://github.com/interledger/web-monetization-extension/pull/534/commits/2c178a023bf5c175caf12b548b4cd7c2e839bf63.

raducristianpopa (approved):

Issue #455, opened 2024-07-29
Monetization tag manager refactoring


raducristianpopa:
### Context

_TBD_

### Todos

_TBD_

PR #538, opened 2024-08-23
refactor(content): simplify MonetizationLinkManager

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/455

- Rename MonetizationTagManager to MonetizationLinkManager (as we want to handle only links in this service)
- Simplify and reorganize MonetizationLinkManager
- TypeScript improvements (check more things)


src/content/services/monetizationLinkManager.ts
```diff
@@ -0,0 +1,498 @@
...
+  }
+
+  /** @throws never throws */
+  private async validateLink(
+    link: HTMLLinkElement,
+  ): Promise<WalletAddress | null> {
+    const walletAddressUrl = link.href.trim();
+    try {
+      checkWalletAddressUrlFormat(walletAddressUrl);
+      const response = await this.message.send('CHECK_WALLET_ADDRESS_URL', {
```
raducristianpopa, line 182: Let's rename `CHECK_WALLET_ADDRESS_URL` to something like `GET_WALLET_ADDRESS_INFO`.
```suggestion
      const response = await this.message.send('CHECK_WALLET_ADDRESS_URL', {
```

src/content/services/monetizationLinkManager.ts
```diff
@@ -0,0 +1,498 @@
...
+    };
+  }
+
+  /** @throws never throws */
+  private async validateLink(
+    link: HTMLLinkElement,
+  ): Promise<WalletAddress | null> {
+    const walletAddressUrl = link.href.trim();
+    try {
+      checkWalletAddressUrlFormat(walletAddressUrl);
```
raducristianpopa, line 181: We can include this function in this class as well if it's not used somewhere else.
```suggestion
      checkHrefFormat(walletAddressUrl);
```

src/content/services/monetizationLinkManager.ts
```diff
@@ -0,0 +1,498 @@
...
+    const monetizationTag = document.querySelector<HTMLLinkElement>(
+      'head link[rel="monetization"]',
+    );
+    return monetizationTag ? [monetizationTag] : [];
+  }
+}
+
+function isNotNull<T>(value: T | null): value is T {
+  return value !== null;
+}
```
raducristianpopa, line 498: Move this into the shared folder.

src/content/services/monetizationLinkManager.ts
```diff
@@ -0,0 +1,498 @@
...
+    return Array.from(
+      document.querySelectorAll<HTMLLinkElement>('link[rel="monetization"]'),
+    );
+  } else {
+    const monetizationTag = document.querySelector<HTMLLinkElement>(
+      'head link[rel="monetization"]',
+    );
+    return monetizationTag ? [monetizationTag] : [];
+  }
+}
```
raducristianpopa, line 494: Let's make this a method of the `LinkManager` class.

src/content/services/frameManager.ts
```diff
@@ -176,21 +176,22 @@ export class FrameManager {
     this.observeDocumentForFrames();
   }
 
+  static handledMessages: ContentToContentMessage['message'][] = [
+    'INITIALIZE_IFRAME',
+    'IS_MONETIZATION_ALLOWED_ON_START',
+    'IS_MONETIZATION_ALLOWED_ON_RESUME',
+  ];
```
raducristianpopa, line 183: What is the reasoning behind making them `static`?

src/content/polyfill.ts
```diff
@@ -1,5 +1,12 @@
 import type { MonetizationEventPayload } from '@/shared/messages';
 (function () {
+  const link = document.createElement('link');
+  if (link.relList.supports('monetization')) {
+    // eslint-disable-next-line no-console
+    console.debug('Monetization is already supported');
+    return;
+  }
```
raducristianpopa, line 8: This shouldn't be needed as this is mostly an issue during development - live reload.

raducristianpopa (commented):
Only code review for now - will have to test the extension as well.

src/content/services/monetizationLinkManager.ts
sidvishnoi: I'd keep such helpers in utils, or as static methods in class, so we can unit test them easily. WDYT @raducristianpopa?

src/content/services/frameManager.ts
sidvishnoi: Such constants make more sense as statics, as they are not depending on anything specific to this class (like class methods too). Given we inject script to each frame, we don't get much of "memory re-use" benefits. So, just here from a theoretical standpoint.

Added to class to keep it _scoped to_ class.

src/content/polyfill.ts
sidvishnoi: This same issue will happen when extension updates from store, and sites might break (or we'll at least get unnecessary errors).

src/content/services/monetizationLinkManager.ts
sidvishnoi: @raducristianpopa Just in case, also rename class from `MonetizationLinkManager` to `LinkManager`? Or too ambiguous?

raducristianpopa: `MonetizationLinkManager` sounds fine to me.

sidvishnoi: @raducristianpopa I think I'll export these functions from this file, and not add them to the class, so we can unit test them. WDYT?

src/content/polyfill.ts
sidvishnoi: Moved context invalidation related fixes out of scope of this PR:
https://github.com/interledger/web-monetization-extension/pull/538/commits/ce72dfbe826491efc7e9a7d1c0d369f8120db612

src/content/services/monetizationLinkManager.ts
raducristianpopa: I don't think that this particular function needs testing. It's mostly a query selector.

src/content/services/frameManager.ts
raducristianpopa: Like you mentioned, since we inject the script in each frame, I would suggest to just make them privatem, not static. This way we avoid doing `FrameManager.handledMessage` and we can just use `this`, which is much shorter.

src/content/services/monetizationLinkManager.ts
raducristianpopa: Sounds good!

raducristianpopa (commented):
It looks like the Monetization event is not sent to links inside iframes nor we send the `START_MONETIZATION` message to background from an iframe?

`load` event is sent on the other hand - the link is being validated.

raducristianpopa (approved):
Before merging this, let's get some feedback from other persons as well to check if this is working as expected.

raducristianpopa (commented):
test-e2e

PR #547, opened 2024-08-28
fix(locales/en): update invalidReceivers error message

jeremiahlee:


sidvishnoi (changes requested):
There was another recommendation from Adam on this same message. I'll find it from meeting notes tomorrow.

sidvishnoi (approved):

PR #549, opened 2024-08-29
fix(background): use default icon on non-http/https tabs

sidvishnoi:
## Context

From Jeremiah's feedback:
> When creating a new tab, which is a page without a wallet address, the bright red icon stole my attention. ... I thought something was wrong or needed my attention.


raducristianpopa (changes requested):
Let's discuss this in tomorrow's call. Please add it as an agenda item.

src/background/services/tabEvents.ts
```diff
@@ -96,18 +97,23 @@ export class TabEvents {
...
     tabId: TabId,
+    tabUrl?: string,
     isTabMonetized: boolean = tabId
       ? this.tabState.isTabMonetized(tabId)
       : false,
     hasTabAllSessionsInvalid: boolean = tabId
       ? this.tabState.tabHasAllSessionsInvalid(tabId)
       : false,
   ) => {
+    tabUrl ||= (await this.browser.tabs.get(tabId)).url;
```
raducristianpopa, line 113: Let's not mutate the function parameter.

raducristianpopa (approved):

src/background/services/tabEvents.ts
sidvishnoi: How do you propose we set a value if not provided in this case? We can't use await in param defaults.

(I'm fine with mutating params, as long as we're not doing something like `param.foo = bar`, I mean, as long as callsite isn't affected)

raducristianpopa: Hmm, do we even need the call to the `tabs` API? On newly created tabs the URL will default to a specific blank page `about:newtab`. What will be the case when the tab URL will not be defined in this case?

PR #582, opened 2024-09-10
chore(renovate.json): increase PR hourly limit

sidvishnoi:
Let it create more PRs.


renovate.json
```diff
@@ -4,7 +4,7 @@
   "extends": ["config:base"],
   "prConcurrentLimit": 10,
   "branchConcurrentLimit": 0,
-  "prHourlyLimit": 5,
+  "prHourlyLimit": 10,
```
raducristianpopa, line 7: I would update the schedule as well to get PRs more frequently. Looking at the dashboard, it looks like there are some pending updates.

sidvishnoi: I think schedule is fine, to avoid noise from dependency updates. We'll do this once a week, not more often.
I'd say let's try this, then we can see if backlog clears fast enough.

DarianM (approved):
once a week schedule is fine, otherwise can get noisy

PR #598, opened 2024-09-16
perf(background/tabEvents): memoize `setIcon` call

sidvishnoi:
In Chrome, we get 3 fetch requests on each `setIcon` calls even if setting same icon. Each request takes 5-25ms, which is unnecessary and noisy as well.

### DIFF
--- a/src/background/services/tabState.ts
+++ b/src/background/services/tabState.ts
@@ -22,6 +22,7 @@ export class TabState {
 
   private state = new Map<TabId, Map<string, State>>();
   private sessions = new Map<TabId, Map<SessionId, PaymentSession>>();
+  private currentIcon = new Map<TabId, Record<number, string>>();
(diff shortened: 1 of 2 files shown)


src/background/services/tabState.ts
```diff
@@ -118,6 +119,14 @@ export class TabState {
     return [...this.sessions.values()].flatMap((s) => [...s.values()]);
   }
 
+  getIcon(tabId: TabId) {
+    return this.currentIcon.get(tabId);
+  }
+
+  setIcon(tabId: TabId, icon: Record<number, string>) {
+    this.currentIcon.set(tabId, icon);
```
sidvishnoi, line 127: Note: we're only storing the reference here, not the full data.

raducristianpopa (approved):

PR #599, opened 2024-09-16
test(e2e): retain traces for failed runs

raducristianpopa:
## Context

With the current setup, if the first run fails we are not saving the trace for it, only for the first retry.

Retain traces only for failed runs.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -26,7 +26,7 @@ export default defineConfig({
       { open: 'never', outputFolder: path.join(testDir, 'playwright-report') },
     ],
   ],
-  use: { trace: 'on-first-retry' },
+  use: { trace: 'retain-on-failure' },
 
   projects: [
     {


raducristianpopa (commented):
test-e2e

playwright.config.ts
```diff
@@ -26,7 +26,7 @@ export default defineConfig({
       { open: 'never', outputFolder: path.join(testDir, 'playwright-report') },
     ],
   ],
-  use: { trace: 'on-first-retry' },
+  use: { trace: 'retain-on-failure' },
```
sidvishnoi, line 29: Or do we want this?
```suggestion
  use: { trace: 'on' },
```

sidvishnoi: No, checked docs. One you added is best one.

raducristianpopa: I don't mind having this. Are we going to use the traces from successful runs as well?

sidvishnoi: No. I just checked docs. retain-on-failure is best.

raducristianpopa (commented):
test-e2e

sidvishnoi (approved):

Issue #590, opened 2024-09-11
[BUG] `HTMLLinkElement.relList.supports('monetization')` is false on page load


sidvishnoi:
### Steps to reproduce

1. Go to any website with extension enabled and following code:
   ```js
   console.log(
      'support?',
      document.createElement('link').relList.supports('monetization'),
   );
   ```

### Expected result

Logs `support? true`

### Actual result

Logs `support? false`

### Additional context

This works fine if we add the check inside a timeout. So the polyfill script is injected a bit late than actual `document_start` as we want it to be.

### Operating system

Linux

### Browsers

Chrome, Firefox

### Extension version

main branch as of today

sidvishnoi:
Tried various ways to load content script before scripts on page, but to no avail. Require webpages to load script with network (instead of inline) doesn't work consistently either.

What works is injecting the polyfill via manifest, and not via content script's `injectPolyfill`. But we require execution `world: "MAIN"` there, which is not well supported in Firefox. Firefox 128 (Released 2024-07-09) adds support for this, but then we lose significant browser support.


So, until Firefox support for world is baseline, we can use that approach. Until then, we've two options:
- Recommend users to check for support in a `setTimeout` (300ms should be enough, but that can depend machine-to-machine)
- Support Chrome with `world: MAIN`, and require timeout support for Firefox (when we know MAIN isn't supported).

sidvishnoi:
As you can see in this video, sometimes it times right and we know the support status on page load. But most of the times, we need a wait to know.

[Screencast from 12-09-24 03:50:54 PM IST.webm](https://github.com/user-attachments/assets/a93f3c16-c24d-480b-836e-378d4b17bd11)

sidvishnoi:
And if we can use `world: "MAIN"` as describe above (Chrome good support, Firefox very recent support), it works consistently:

[Screencast from 12-09-24 03:53:24 PM IST.webm](https://github.com/user-attachments/assets/47e9c3ad-5f6d-4b63-9a45-607b6d754e6a)

sidvishnoi:
cc @raducristianpopa

raducristianpopa:
Can you please include this in the agenda for this week's call? I am okay with sacrificing some Firefox browser support in exchange of a consistent way, that works the same in every browser, to check if Web Monetization is supported.

sidvishnoi:
This is what happens in Firefox 127 when we inject polyfill via unsupport `world: MAIN` only:
[...]

PR #606, opened 2024-09-17
fix: inject polyfill earlier when `world: MAIN` is supported

sidvishnoi:
## Context

Fixes https://github.com/interledger/web-monetization-extension/issues/590
Without needing to sacrifice Firefox support!

If browser supports `world: MAIN` for content script, we inject directly using `scripting.registerContentScripts` API (can detect WM support immediately on page load). If not, we inject like before from contentScript (slow, requires timeout for WM support detection).

We inject via `scripting.registerContentScripts` API and not directly via `manifest.json`, as we can catch the lack of support of `world: MAIN` via this API. With this detection, we can still support Firefox versions older than 128.

Update polyfill to support injecting twice - only the first injection works - as we detect if already injected.

Note: In content script, we inject polyfill from contentScript and not using `scripting.executeScript` in background as we'd otherwise need to inject it manually to each tab.

### What next?

When Firefox 128 is old enough, we can remove `injectPolyfill` from background.ts as well as `contentScript.ts` and register it directly in manifest.json with `world: MAIN`. When we do that, we'd update `browser_specific_settings.gecko.strict_min_version` in manifest.json from 110 to 128, and also remove `polyfill/*` from `web_accessible_resources`.

### Test results:

| Browser | WM detection on page load | WM detection later | Other JS Events |
| --- | :---: | :---: | :---: |
| Firefox 126 | ❌ | ✅ | ✅ |
| Firefox 130 | ✅ | ✅ | ✅ |
| Chrome 110 |  ✅ | ✅ | ✅ |

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/ExecutionWorld#browser_compatibility


sidvishnoi (commented):
test-e2e

src/background/services/background.ts
```diff
@@ -61,6 +62,30 @@ export class Background {
     this.sendToPopup.start();
   }
 
+  async injectPolyfill() {
+    try {
+      // TODO: When Firefox 128 is old enough, inject directly via manifest.
+      // Also see: injectPolyfill in contentScript
```
raducristianpopa, line 68: Can you please log an issue for this? Please add the link for the issue in the comment as well.

src/background/services/background.ts
```diff
@@ -61,6 +62,30 @@ export class Background {
...
+          matches: PERMISSION_HOSTS.origins,
+          runAt: 'document_start',
+        },
+      ]);
+    } catch (error) {
+      // Firefox <128 will throw saying world: MAIN isn't supported. So, we'll
+      // inject via contentScript later. Injection via contentScript is slow,
+      // but apart from WM detection on page-load, everything else works fine.
+      if (!error.message.includes(`world`)) {
+        this.logger.error(error);
```
raducristianpopa, line 84: Lets add more context for this error. I'm thinking about something along these lines:

```
World `MAIN` not supported by your browser - {{ browser.details }}. Upgrading to Firefox version 128 is going fix this error. 

Error: {{ error }}
```

I assume that if someone is looking in the DevTools they will probably understand why this is happening.

src/content/polyfill.ts
```diff
@@ -1,5 +1,10 @@
 import type { MonetizationEventPayload } from '@/shared/messages';
 (function () {
+  if (document.createElement('link').relList.supports('monetization')) {
+    // console.log('already patched');
```
raducristianpopa, line 4: We can remove this log.

sidvishnoi:
> The timing issue for Firefox < 128 still exists right?

Yes, but it'll at least continue to work like before. We just won't need to set min supported version to 128 this way.

src/content/polyfill.ts
sidvishnoi: Ok, but I'll add the comment with same message. So we know what we're doing here.

raducristianpopa (approved):
We can still discuss in this week's call if we want to drop support for older Firefox versions.

Issue #67, opened 2023-12-20
WM ext browser support - Edge


tselit:
### Context

WM extension compatibility with the Microsoft Edge browser.

### Todos

TBC

tselit:
Bibi to do: test on Edge and then should be ready to close

sidvishnoi:
To test: How it works in split-window mode (https://github.com/interledger/web-monetization-extension/pull/366#discussion_r1673836423)

raducristianpopa:
In split view, the extension is only taking into the consideration only the first pane from the split view when updating icon. If we are on a non-monetized website in the first pane and on a web monetized one in the second one, the icon will have the non-monetized website identifier.

<img width="990" alt="image" src="https://github.com/user-attachments/assets/11b573db-4a92-4ca2-88df-085f3bef780f">

raducristianpopa:
Arc's split view should be checked as well.

RabebOthmani:
Blocked as we wait for the MS support team

sidvishnoi:
https://github.com/microsoft/MicrosoftEdge-Extensions/discussions/102 🤷🏽‍♂️🤦🏽

sidvishnoi:
Found the sender port in Runtime.Port message handler returns correct tab 🎉 
```
port.onMessage.addListener((message, sendingPort) => {
   console.log(sendingPort.sender?.tab?.id);
});
```

So, we can make a short connection on content script inject and store tabId for use in future messages?

sidvishnoi:
Everything works once we get right tabId, except Edge doesn't update the icon (even with correct tabId!)

Issue #146, opened 2024-03-07
[NEW FEATURE] Requirements - Exceptions List - add a site-specific rate of pay


raducristianpopa:
### Context

As a WM extension user, I want to configure site-specific hourly rates such that I can have a different rate of pay for each website.

### Describe the solution you would like

A way for users to configure and maintain a list of site-specific hourly rates of pay, in the context of continuous payments, where the minimum hourly rate can be 0, and the maximum 1.

### Additional context

Some things to consider: 
- config hierarchy such that config for a path to a specific resource within a domain supercedes config for the domain (e.g. config for `www.furryfriends.org/bobtail` supercedes config for `www.furryfriends.org`)
- users may want to perform bulk actions to multiple websites at the same time

### Todos

- [ ] Add debounce for the rate of pay slider;
- [ ] After changing the rate of pay, send the updated amount to the background script, convert the amount to the asset scale of the wallet (represented as a string, eg: asset code USD, asset scale 2, amount 0.01 -> `amount * 10 ** scale`) and add the website in the exception list using the below format;
- [ ] If the website is already in the exception list, only update the amount (after converting it as well);

#### Format
```json
{
  "exceptionList": {
    "https://example.com": {
      "amount": "60", 
      "interval": "3600000" // in ms
    }
  }
}
```
The `interval` property is for future proofing. For now, we will only store the default value for the interval, 1 hour in milliseconds.

PR #608, opened 2024-09-17
feat: support split-view in Edge, tab-tiling in Vivaldi

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/67

Why:
- In Edge, `browser.tabs.query()` skips tabs in right-pane of split-view, as if it don't exist. So, we need to keep track of existing tabs in a window ourselves.
- Similarly, we can't know which tab was last focused.
- There's no API to tell which all tabs are visible (that can be monetized) in a Edge split-view or Vivaldi tab-tiling view.

- Fire a `TAB_FOCUSED` event from content script, on load (if focused) and on focus change (if focused).
- Use the `sender.tab.id` from above event to store last focused tabId for given window.
- Store that info in newly added `WindowState` service.
- Also store last focused window in above service.
- Note: Both pages in split-mode are monetized together.
- For Edge, if we call `browser.action.setIcon()` with a tabId, it'll only update the icon for the left pane. So, added a `browserName` value, and skip using `tabId` for Edge when setting icon. This will set a global icon for Edge, but other browsers stay like before.
- Instead of relying on `onActivatedTab` which is buggy in Edge with split-view, rely on `TAB_FOCUSED` events' `sender.tab.id` to update icon state. Also use this last focused tab to show popup content.
- In order to support start/stop monetizing all tabs in given view on window focus change (not same as `visibilitychange`, which works like before), we send `IS_TAB_IN_VIEW` message to content script (top frame only) to know if tab is currently visible. This way, we get all visible tabs in current window.
- Remove `IS_WM_ENABLED` message as unused.
- Made some improvements to avoid needless checking need to pause/resume on focus change when extension popup is opened.

For reviewers: Radu is on vacation so I was asked to ask you good folks instead.

## To be decided

Right now, the rate of pay is decided per tab, using the global rate of pay. Now we have multiple active visible tabs getting monetized, so the effective global rate of pay is multiplied by number of active visible tabs - user is paying more than they initially set to as they're paying multiple websites at once. In future, we'll have exception lists (https://github.com/interledger/web-monetization-extension/issues/146), which will again make rate of pay per site/tab.

### DIFF
--- a/src/background/container.ts
+++ b/src/background/container.ts
@@ -7,6 +7,7 @@ import {
   Background,
   TabEvents,
   TabState,
+  WindowState,
   SendToPopup,
   EventsService,
   Heartbeat,
@@ -14,7 +15,12 @@ import {
 } from './services';
 import { createLogger, Logger } from '@/shared/logger';
 import { LOG_LEVEL } from '@/shared/defines';
-import { tFactory, type Translation } from '@/shared/helpers';
+import {
+  getBrowserName,
+  tFactory,
+  type BrowserName,
+  type Translation,
+} from '@/shared/helpers';
 import {
   MessageManager,
   type BackgroundToContentMessage,
@@ -23,6 +29,7 @@ import {
 export interface Cradle {
   logger: Logger;
   browser: Browser;
+  browserName: BrowserName;
   events: EventsService;
   deduplicator: Deduplicator;
   storage: StorageService;
@@ -34,6 +41,7 @@ export interface Cradle {
   background: Background;
   t: Translation;
   tabState: TabState;
+  windowState: WindowState;
   heartbeat: Heartbeat;
 }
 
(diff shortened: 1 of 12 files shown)


sidvishnoi:
If we ignore `tabId` when setting icon in Edge, icon update works fine too. But as it's inefficient, we'll do it only for Edge.

[Screencast from 17-09-24 07:48:39 PM IST.webm](https://github.com/user-attachments/assets/29da469b-e82a-4be8-a4a5-aa4b1e10cd45)

sidvishnoi:
TODO: handle start/resume for all open tabs in split mode, not just left pane.

sidvishnoi:
Demo with Vivaldi tab-tiling!

[Screencast from 18-09-24 06:53:57 PM IST.webm](https://github.com/user-attachments/assets/2dad11ab-bbcc-440e-9e66-4b6df1f70a4a)

Tymmmy (approved):
While I wasn't able to actually run this right now to see the split view on Edge (nothing to do with you, but with my local current setup), the code looks good to me.

Tymmmy:
@sidvishnoi I have to ask, is this story ready to review? Since I reviewed it and gave approval, you pushed at least 5 new commits. I am asking so that we know when to actually review?

Tymmmy (approved):
LGTM, surprised we support Edge :)

lengyel-arpad85 (approved):

Issue #595, opened 2024-09-13
[BUG] Extension unable to disconnect if grant given when connecting has since been revoked


tselit:
Related to #332 

### Steps to reproduce
1. Using rafiki.money, started with a USD account balance of about $41.74.
2. Connected the WM extension to the account's wallet address, with a budget of $20.  <br>
Then, I did the following using [rafiki.money](https://rafiki.money):
3. I manually tried to withdraw two x $20 from the wallet, but the wallet balance appeared to remained unchanged at $41.74.
4. Successfully transferred $1.02 to another payment pointer in a different wallet (note: $0.02 was the fee).
5. The new wallet balance was $41.72.
6. I manually went to revoke the $20 grant that had been issued when the extension was connected.
7. Observed that the account balance was immediately reduced by $40 (I believe this reflects the two x $20 withdrawals that had not succeeded earlier, in step 3). 
8. The new wallet balance was $0.72

9. Navigate to the WM playground which still had 2 receiving payment pointers that had been actively receiving payments earlier.
10. On the WM playground, use the WM extension to send a one-time payment of $0.20.
    -- **Result**: The extension displayed the message "Could not facilitate payment for current website"
11. Attempted to disconnect the extension from the wallet.
    -- **Result**: The extension remained connected, and was unable to disconnect. While on the WM playground with the 2 payment pointers, the extension icon still appears active with a green tick.
12. Navigated back to rafiki.money to remove the WM extension's public key from the payment pointer that had been used to connect to the wallet.
13. Attempted to disconnect the extension from the wallet.
    -- **Result**: The extension remained connected, and was unable to disconnect. While on the WM playground with the 2 payment pointers, the extension icon still appears active with a green tick.

### Expected result

- After step 11 above, I had expected the extension to disconnect from the wallet successfully.  

**Something to consider:**
- After step 10 above, should the extension indicate a different message to indicate that the extension actually no longer has access/permissions, or is the "Could not facilitate payment" message still suitable for this context?

### Actual result

- After step 10, the **result**: The extension displayed the message "Could not facilitate payment for current website"
- After step 11, the **result**: The extension remained connected, and was unable to disconnect. While on the WM playground with the 2 payment pointers, the extension icon still appears active with a green tick.
- After step 13, the **result**: The extension remained connected, and was unable to disconnect. While on the WM playground with the 2 payment pointers, the extension icon still appears active with a green tick.

### Screenshots or videos

[...]

PR #609, opened 2024-09-17
fix(bg/openPayments): handle `invalid_continuation` on disconnect

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/595

### DIFF
--- a/src/background/services/openPayments.ts
+++ b/src/background/services/openPayments.ts
@@ -628,7 +628,7 @@ export class OpenPaymentsService {
     try {
       await this.client!.grant.cancel(grantContinuation);
     } catch (error) {
-      if (isInvalidClientError(error)) {
+      if (isInvalidClientError(error) || isInvalidContinuationError(error)) {
         // key already removed from wallet
         return;
       }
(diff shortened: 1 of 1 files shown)


src/background/services/openPayments.ts
```diff
@@ -787,6 +787,11 @@ export const isInvalidClientError = (error: any) => {
   return error.status === 400 && error.code === 'invalid_client';
 };
 
+export const isInvalidContinuationError = (error: any) => {
+  if (!isOpenPaymentsClientError(error)) return false;
+  return error.status === 401 && error.code === 'invalid_continuation';
```
mkurapov, line 792: Based on what I see in the code, when revoking/cancelling grant, we should not end up returning this error. 
`invalid_continuation` only happens on grant continuation

sidvishnoi: It is infact happening as seen here: https://github.com/interledger/web-monetization-extension/issues/595#issuecomment-2355760034 when trying to cancel grant due to `getByContinue` being used here:

https://github.com/interledger/rafiki/blob/9a7eed9114ac9cc4cd412e6d5e8c8af7b3879a55/packages/auth/src/signature/middleware.ts#L90-L101
+
https://github.com/interledger/rafiki/blob/9a7eed9114ac9cc4cd412e6d5e8c8af7b3879a55/packages/auth/src/app.ts#L288-L297

mkurapov: Excuses, yes, you are correct :)

mkurapov (approved):

Issue #560, opened 2024-09-02
Pass latest state when page navigates while popup is open


sidvishnoi:
1. Visit a monetized website.
2. Run following to console to simulate background navigation (or a slow navigation - user clicks,  user opens popup, site thinking, site navigates):
    ```js
    setTimeout(() => {window.location.href = "https://another-monetized-website"}, 4000) 
    ```
3. Open popup before above setTimeout callback is called.
4. Wait for site to be navigated.

The popup still has state from previous site.

This doesn't happen too often in real world, but handling this correctly will help us with https://github.com/interledger/web-monetization-extension/pull/556#issuecomment-2324809958.

Issue #584, opened 2024-09-10
[IMPROVEMENT] Show msg that WM ext only supports HTTPS sites


tselit:
### Current behaviour

When visiting a monetized website that has a valid link tag, **but** where the website does not have an SSL certificate (e.g. the URL doesn't have an HTTPS prefix or the page's SSL certificate has expired), then WM extension displays this message "This website is not monetized".

### Describe the feature you would like to request

When a web site or page is monetized but it doesn't have a valid SSL certificate (i.e. the URL doesn't have the HTTPS prefix), then let the extension display a more informative and accurate message, to indicate that the extension does not work on site/page.

### To do

When a user of the extension visits a website or page that isn't HTTPS compliant, do the following:
- Show the active icon of the extension.
- When opening the extension, show a message that Web Monetization only works on secure sites.

### Acceptance criteria

When a user navigates to a non-HTTPS compliant website:
1. The icon of the extension should appear as the default, active icon (i.e. without any green tick, or a red X, or an orange exclamation point). 
![default](https://github.com/user-attachments/assets/a6b67ac0-4b4f-4946-bbfb-e2c54b2d9868)

2. Show this message when the user opens the extension: `Unable to start Web Monetization. Web Monetization only works on secure sites.`

3. For all non-HTTPS compliant sites or pages, the above points 1 and 2 should always apply, regardless of the monetization status.

### Additional context

Related to issue logged #572

sidvishnoi:
Would like if the messaging reflects that we cannot monetized empty tabs, `about:`/`chrome://`/`file:///` (local files) pages as well.

RabebOthmani:
How about : This is a non supported page. For more details about the Web Monetization usage, check out [link]' 
This way it's generic enough to cover the different cases you mentioned @sidvishnoi and we can add the explanation in the getting started doc ? cc @tselit @melissahenderson

tselit:
Cool, this sounds good to me @RabebOthmani.

melissahenderson:
Is this messaging for testers or is it what we're planning on having in v1? Including "for more details...about usage..." is for developers, not your average user. The average user can't do anything to make a site secure.

In any case, a phrasing suggestion I have is: Unable to start Web Monetization. Web Monetization only works on secure sites. 

I don't particularly like "start", but something along those lines.

RabebOthmani:
[...]

PR #611, opened 2024-09-19
fix: update popup tab data along with icon

sidvishnoi:
- Closes https://github.com/interledger/web-monetization-extension/issues/560
- Will help with https://github.com/interledger/web-monetization-extension/issues/584

- refactor(PopupState): add all popup data related to tab under `tab`
   - `tabUrl` -> `tab.url`
   - `isSiteMonetized` -> `tab.status === 'monetized'`
   - `hasAllSessionsInvalid` -> `tab.status === 'all_sessions_invalid'`
- refactor(monetization): extract `getPopupTabData` to tabState service
- refactor(tabEvents): use `getPopupTabData` instead of multiple params

Paves way to show more useful messages with `tab.status` (e.g. new tab, internal extension pages, non-https pages)

### DIFF
--- a/src/background/services/tabEvents.ts
+++ b/src/background/services/tabEvents.ts
@@ -1,6 +1,5 @@
 import { isOkState, removeQueryParams } from '@/shared/helpers';
-import { ALLOWED_PROTOCOLS } from '@/shared/defines';
-import type { Storage, TabId } from '@/shared/types';
+import type { PopupTabInfo, Storage, TabId } from '@/shared/types';
 import type { Browser, Tabs } from 'webextension-polyfill';
 import type { Cradle } from '@/background/container';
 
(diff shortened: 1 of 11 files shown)


src/background/services/tabEvents.ts
```diff
@@ -187,26 +171,28 @@ export class TabEvents {
...
-    } else if (!isOkState(state) || hasTabAllSessionsInvalid) {
+    } else if (!isOkState(state) || tabInfo.status === 'all_sessions_invalid') {
       iconData = enabled ? ICONS.enabled_warn : ICONS.disabled_warn;
       const tabStateText = this.t('icon_state_actionRequired');
       title = `${title} - ${tabStateText}`;
+    } else if (
+      tabInfo.status !== 'monetized' &&
+      tabInfo.status !== 'no_monetization_links'
+    ) {
+      // use defaults
```
lengyel-arpad85, line 193: can't this be formulated in a way so we don't have an empty code block, then an else branch ? so an if without an else that actually executes some code

sidvishnoi: Right now it's a no-op, but with https://github.com/interledger/web-monetization-extension/issues/584, we're likely going to change the icons. Kept this flow to keep space when we need to change icons, and also to keep the diff small.

lengyel-arpad85 (approved):

Issue #418, opened 2024-07-11
Handle extension context invalidation


sidvishnoi:
When extension is updated/disabled/enabled, the content scripts (including polyfill) keep connections with old background script, which leads to "Error: Extension context invalidated" in browser console.
We need to detect the disconnect/invalidation, and tear-down/reset content scripts correctly.

RabebOthmani:
@sidvishnoi I see the 2 PRs merged. Can I assume this is done?

sidvishnoi:
@RabebOthmani No. The PRs relate to it, but don't complete it.

sidvishnoi:
@RabebOthmani It's very specific to Chrome and the de-facto way requires all pages to be reloaded in browser - which can lead to user losing their data if done automatically.
One way is make action be sent to "action required" state and ask user to refresh the tab for extension to continue working.

So, there isn't really a good way to fix it - just gradual improvements that reduce chances of this happening (as the PRs related to it try to do).

That's why the issue is open. If anyone else reports, we can direct them here - that we know it's an issue that we know.

PR #633, opened 2024-09-27
chore(scripts/dev): reload page on content script change

sidvishnoi:
## Context

Related to https://github.com/interledger/web-monetization-extension/issues/418, but only for dev.

Avoids need to manually reload every time there's change in content scripts or background (which reloads extension, hence content scripts).

### DIFF
--- a/esbuild/dev.ts
+++ b/esbuild/dev.ts
@@ -99,7 +99,7 @@ function liveReloadPlugin({ target }: { target: Target }): ESBuildPlugin {
         const contents = await readFile(args.path, 'utf8');
         return {
           contents: contents + '\n\n\n' + reloadScriptContent,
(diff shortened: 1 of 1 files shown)


esbuild/dev.ts
```diff
@@ -83,6 +95,13 @@ function liveReloadPlugin({ target }: { target: Target }): ESBuildPlugin {
           loader: 'tsx' as const,
         };
       });
+      build.onLoad({ filter: /src\/content\// }, async (args) => {
+        const contents = await readFile(args.path, 'utf8');
+        return {
+          contents: contents + '\n\n\n' + reloadScriptContent,
+          loader: 'tsx' as const,
```
sidvishnoi, line 102:
```suggestion
          loader: 'ts' as const,
```

Issue #613, opened 2024-09-20
[META] Setup page improvements


sidvishnoi:
- [x] #614
   - [x] Make copy updates and validation improvements based on https://github.com/interledger/web-monetization-extension/issues/565
   - [x] Support fallback for copy-paste key when automatic copying key fails
- [x] Add service to automatically copy public key to the wallet  / along with E2E tests for each ([related](https://github.com/interledger/web-monetization-extension/issues/311))
   - [x] #615
   - [x] #616
   - [x] #617
- [x] #625
- [x] #622

Out of scope for this story:
- #618 (depends on GateHub fixing some things)
- #646
- #653

PR #638, opened 2024-10-02
fix(ConnectWalletForm): preserve errors on popup reopen

sidvishnoi:
## Context

Part of https://github.com/interledger/web-monetization-extension/issues/613

[Screencast from 02-10-24 05:45:59 PM IST.webm](https://github.com/user-attachments/assets/a6620bac-439a-4cfc-a6f0-15b4c243e900)

### DIFF
--- a/src/background/services/background.ts
+++ b/src/background/services/background.ts
@@ -194,7 +194,7 @@ export class Background {
 
             case 'CONNECT_WALLET':
               await this.openPaymentsService.connectWallet(message.payload);
-              if (message.payload.recurring) {
+              if (message.payload?.recurring) {
                 this.scheduleResetOutOfFundsState();
               }
               return success(undefined);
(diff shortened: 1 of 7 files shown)


raducristianpopa (commented):
A little bit confused. So we will preserve the error even if the popup is opened multiple times?

sidvishnoi:
Yes, so user knows what went wrong. For example, we open new tab, popup closes:
- user closes tab, then we can show "you closed tab too soon" error message.
- key add fails, user opens popup again, they connect again - we end up trying to add key over and over, as we lost context on popup reload.

We need that context (even if we don't show error message). These will be cleared when user successfully connects (or resets when user tries to connect again, to be replaced with newer error message); or when browser is closed.

sidvishnoi:
TODO: reset all error and state when wallet address is changed.

raducristianpopa (approved):

PR #640, opened 2024-10-03
fix(content/KeyAutoAdd): get output of any step, not just of previous

sidvishnoi:
Simplifies KeyAutoAdd's step runner to let us get output of any step (that has been previously run), not just result of previous. Getting result of previous had limitation that any intermediate steps had to pass through data from previous steps. Now we store outputs in a Map and get those with the `output(fn)` helper (WeakMap not needed, as we store reference to function anyway in steps array).

Also update signature of StepRun function.

Additionally, turn `skip` call to throw an instance of `SkipError` instead of object with symbol.

### DIFF
--- a/src/content/keyAutoAdd/lib/keyAutoAdd.ts
+++ b/src/content/keyAutoAdd/lib/keyAutoAdd.ts
@@ -43,7 +43,7 @@ export class KeyAutoAdd {
     this.port.onMessage.addListener(
       (message: BackgroundToKeyAutoAddMessage) => {
         if (message.action === 'BEGIN') {
-          this.run(message.payload);
+          this.runAll(message.payload);
         }
       },
     );
@@ -121,7 +121,7 @@ export class KeyAutoAdd {
     return promise;
   }
 
-  private async run(payload: BeginPayload) {
+  private async runAll(payload: BeginPayload) {
     const helpers: StepRunHelpers = {
       output: <T extends StepRun>(fn: T) => {
         if (!this.outputs.has(fn)) {
--- a/src/content/keyAutoAdd/lib/keyAutoAdd.ts
+++ b/src/content/keyAutoAdd/lib/keyAutoAdd.ts
@@ -23,8 +23,6 @@ export type { StepRun } from './types';
 
 export const LOGIN_WAIT_TIMEOUT = 10 * 60 * 1000;
 
-const SYMBOL_SKIP = Symbol.for('skip');
-
 export class KeyAutoAdd {
   private port: Runtime.Port;
   private ui: HTMLIFrameElement;
@@ -135,10 +133,9 @@ export class KeyAutoAdd {
       keyId,
       keyAddUrl,
       skip: (details) => {
-        throw {
-          type: SYMBOL_SKIP,
-          details: typeof details === 'string' ? new Error(details) : details,
-        };
+        throw new SkipError(
+          typeof details === 'string' ? { message: details } : details,
+        );
       },
       setNotificationSize: (size: 'notification' | 'fullscreen') => {
         this.setNotificationSize(size);
@@ -165,14 +162,12 @@ export class KeyAutoAdd {
         this.setStatus(stepIdx, 'success', {});
         prevStepId = step.name;
       } catch (error) {
-        if (this.isSkip(error)) {
-          const details = this.errorToDetails(
-            error.details.error || error.details,
-          );
+        if (error instanceof SkipError) {
+          const details = error.toJSON();
           this.setStatus(stepIdx, 'skipped', { details });
           continue;
         }
-        const details = this.errorToDetails(error);
+        const details = errorToDetails(error);
         this.setStatus(stepIdx, 'error', { details: details });
         this.postMessage('ERROR', { details, stepName: step.name, stepIdx });
         this.port.disconnect();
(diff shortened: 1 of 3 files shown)


src/content/keyAutoAdd/testWallet.ts
```diff
@@ -88,16 +85,13 @@ const getAccountDetails: Step<never, Account[]> = async ({
  * with a different account (user disconnected and changed account), revoke from
  * there first.
  */
-const revokeExistingKey: Step<typeof getAccountDetails, Account[]> = async (
-  { keyId, skip },
-  accounts,
-) => {
+const revokeExistingKey: Run<void> = async ({ keyId }, { skip, output }) => {
+  const accounts = output(getAccounts);
```
sidvishnoi, line 89: That's the main thing.

src/content/keyAutoAdd/testWallet.ts
```diff
@@ -106,10 +100,11 @@ const revokeExistingKey: Step<typeof getAccountDetails, Account[]> = async (
...
 
-const findWallet: Step<
-  typeof revokeExistingKey,
-  { accountId: string; walletId: string }
-> = async ({ walletAddressUrl }, accounts) => {
+const findWallet: Run<{ accountId: string; walletId: string }> = async (
+  { walletAddressUrl },
+  { output },
+) => {
+  const accounts = output(getAccounts);
```
sidvishnoi, line 107: See how it doesn't rely anymore on `revokeExistingKey`

sidvishnoi (commented):
test-e2e

src/content/keyAutoAdd/lib/keyAutoAdd.ts
```diff
@@ -121,24 +121,19 @@ export class KeyAutoAdd {
...
+        return this.outputs.get(fn) as Awaited<ReturnType<T>>;
+      },
       skip: (details) => {
-        throw {
-          type: SYMBOL_SKIP,
-          details: typeof details === 'string' ? new Error(details) : details,
-        };
+        throw new SkipError(
+          typeof details === 'string' ? { message: details } : details,
+        );
```
raducristianpopa, line 136: :+1:

src/content/keyAutoAdd/lib/keyAutoAdd.ts
```diff
@@ -121,24 +121,19 @@ export class KeyAutoAdd {
...
-    keyAddUrl,
-  }: BeginPayload) {
-    const params: StepRunParams = {
-      walletAddressUrl,
-      publicKey,
-      nickName,
-      keyId,
-      keyAddUrl,
+  private async run(payload: BeginPayload) {
+    const helpers: StepRunHelpers = {
```
raducristianpopa, line 125: Can `helpers` be a private property of the class, since it doesn't depend on anything? This way we are not going to create this variable every time we call `run` throughout the auto-add flow.

src/content/keyAutoAdd/lib/keyAutoAdd.ts
```diff
@@ -121,24 +121,19 @@ export class KeyAutoAdd {
...
+      output: <T extends StepRun>(fn: T) => {
+        if (!this.outputs.has(fn)) {
```
raducristianpopa, line 127: Just for my understanding, all steps will have an output?

src/content/keyAutoAdd/lib/types.ts
```diff
@@ -1,21 +1,20 @@
...
+  output: <T extends StepRun>(fn: T) => Awaited<ReturnType<T>>;
 }
 
-export type StepRun<T = unknown, R = void> = (
-  params: StepRunParams,
-  prevStepResult: T extends (...args: any[]) => PromiseLike<any>
-    ? Exclude<Awaited<ReturnType<T>>, void | { type: symbol }>
-    : T,
-) => Promise<R | void>;
+export type StepRun<T = unknown> = (
```
raducristianpopa, line 10: I assume that `T` is the output of the step, right? If that is correct let's rename the generic to `TOutput`.

src/content/keyAutoAdd/testWallet.ts
raducristianpopa: The comment for this function is confusing me a bit. If the user disconnected and changed the account, we wouldn't be able to see the key in the keys page.

src/content/keyAutoAdd/lib/keyAutoAdd.ts
sidvishnoi: Run is called only on page reload kinda.

src/content/keyAutoAdd/lib/keyAutoAdd.ts
sidvishnoi: - All steps will have output, unless they throw (error or skip).
- If they return nothing, output is `undefined`, but still exists.
- Only reason `this.outputs.has(fn)` can be false is when the step hasn't run yet, or was skipped/errored.

src/content/keyAutoAdd/testWallet.ts
sidvishnoi: Say, user connected once to `USD -> Addr#1`. Then disconnected. The key is still there in wallet added to `USD -> Addr#1` account.
If now I want to connect my `EUR -> Addr#1` account, the extension still has the same key. So adding it will throw an `internal server error`. That's why we first revoke existing key (by ID) if any (from any existing account/address).

src/content/keyAutoAdd/lib/keyAutoAdd.ts
raducristianpopa: Can't we populate the outputs from the beginning and have `undefined` for no output yet and something else if they return nothing - like an empty object?

src/content/keyAutoAdd/testWallet.ts
raducristianpopa: Oh, I thought you were referring to a Test Wallet account in the comment. Ok, makes sense now. Let's add a TODO for this since we will probably won't have to do this in the future: users should be able to upload the same key to multiple wallet addresses, but not to the same one.

raducristianpopa (commented):
A small additional comment: I would suggest adding the base URLs in environment variables. It would be easier to update them when the Test Wallet is going to move to the new domain.

src/content/keyAutoAdd/lib/keyAutoAdd.ts
sidvishnoi: So, the way `output(fn)` is supposed to be called, is, it has to make sure that output you're trying to get infact exists. For example, `waitForLogin` must not call `output(addKey)`. This throwing will ensure we catch such bugs in dev only.

If they return nothing, `output(fn)` will return `undefined` (and not throw).

src/content/keyAutoAdd/lib/keyAutoAdd.ts
raducristianpopa: The other function called `run` in this method threw me off.

raducristianpopa (approved):

sidvishnoi:
> I would suggest adding the base URLs in environment variables. It would be easier to update them when the Test Wallet is going to move to the new domain.

I thought of doing that, until I found there's `api.rafiki.money` too, then thought hard-coding is fine. Might revisit as I work on Fynbos one.

PR #637, opened 2024-10-02
feat(ConnectWalletForm): ask consent for automatic key addition

sidvishnoi:
## Context

- Closes https://github.com/interledger/web-monetization-extension/issues/625
- Part of https://github.com/interledger/web-monetization-extension/issues/613

## Changes proposed in this pull request

[Screencast from 02-10-24 09:22:43 PM IST.webm](https://github.com/user-attachments/assets/53317616-4608-4fc8-b7ca-16529a65159d)

PR #641, opened 2024-10-03
refactor(ConnectWalletForm): preserve key in errors

sidvishnoi:
Earlier, we stored only localized message in `errors[foo]`. Now we preserve the `key` and original error under `{info}`; and add the localized message under `{message}`.

This will help with error handling based on key, and paves way to support retry add-key on retry-able errors (such as timed out waiting for login, wrong account was logged in, key-add consent accidentally declined etc.)

Part of https://github.com/interledger/web-monetization-extension/issues/613

### DIFF
--- a/src/popup/components/ConnectWalletForm.tsx
+++ b/src/popup/components/ConnectWalletForm.tsx
@@ -162,8 +162,8 @@ export const ConnectWalletForm = ({
     [saveValue, currencySymbol, toErrorInfo],
   );
 
-  const handleSubmit = async (ev?: React.FormEvent<HTMLFormElement>) => {
-    ev?.preventDefault();
+  const handleSubmit = async (ev: React.FormEvent<HTMLFormElement>) => {
+    ev.preventDefault();
 
     const errWalletAddressUrl = validateWalletAddressUrl(walletAddressUrl);
     const errAmount = validateAmount(amount, currencySymbol.symbol);
--- a/src/popup/components/ConnectWalletForm.tsx
+++ b/src/popup/components/ConnectWalletForm.tsx
@@ -72,7 +72,7 @@ export const ConnectWalletForm = ({
 
   const resetState = React.useCallback(async () => {
     await clearConnectState();
-    setErrors((_) => ({ ..._, keyPair: null, connect: null }));
+    setErrors((prev) => ({ ...prev, keyPair: null, connect: null }));
     setAutoKeyShareFailed(false);
   }, [clearConnectState]);
 
@@ -109,7 +109,7 @@ export const ConnectWalletForm = ({
 
   const getWalletInformation = React.useCallback(
     async (walletAddressUrl: string): Promise<boolean> => {
-      setErrors((_) => ({ ..._, walletAddressUrl: null }));
+      setErrors((prev) => ({ ...prev, walletAddressUrl: null }));
       if (!walletAddressUrl) return false;
       try {
         setIsValidating((_) => ({ ..._, walletAddressUrl: true }));
@@ -117,8 +117,8 @@ export const ConnectWalletForm = ({
         const walletAddress = await getWalletInfo(url.toString());
         setWalletAddressInfo(walletAddress);
       } catch (error) {
-        setErrors((_) => ({
-          ..._,
+        setErrors((prev) => ({
+          ...prev,
           walletAddressUrl: toErrorInfo(error.message),
         }));
         return false;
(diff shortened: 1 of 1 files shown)


sidvishnoi:
image

src/popup/components/ConnectWalletForm.tsx
```diff
@@ -62,34 +65,34 @@ export const ConnectWalletForm = ({
...
   );
+
   const [autoKeyShareFailed, setAutoKeyShareFailed] = React.useState(
     isAutoKeyAddFailed(state),
   );
 
   const resetState = React.useCallback(async () => {
     await clearConnectState();
-    setErrors((_) => ({ ..._, keyPair: '', connect: '' }));
+    setErrors((_) => ({ ..._, keyPair: null, connect: null }));
```
raducristianpopa, line 75: I suggest renaming the parameter to `prev` or something else that suggests that this is the previous state. Clearer than `_`.
```suggestion
    setErrors((prev) => ({ ...prev, keyPair: null, connect: null }));
```

src/popup/components/ConnectWalletForm.tsx
```diff
@@ -62,34 +65,34 @@ export const ConnectWalletForm = ({
...
     setAutoKeyShareFailed(false);
   }, [clearConnectState]);
 
+  const toErrorInfo = React.useCallback(
+    (err?: string | ErrorWithKeyLike | null): ErrorInfo | null => {
```
raducristianpopa, line 80: Why do we have the error as string as well? Can't we pass the whole error directly? For example here:

https://github.com/interledger/web-monetization-extension/blob/a012ce4ec04def3ee3a9400eab2f3b2411246f72/src/popup/components/ConnectWalletForm.tsx#L114-L124

src/popup/components/ConnectWalletForm.tsx
```diff
@@ -62,34 +65,34 @@ export const ConnectWalletForm = ({
...
-      state?.status === 'error'
-        ? isErrorWithKey(state.error)
-          ? t(state.error)
-          : state.error
-        : '',
+  const [errors, setErrors] = React.useState<Errors>({
+    walletAddressUrl: null,
+    amount: null,
+    keyPair: state?.status === 'error:key' ? toErrorInfo(state.error) : null,
+    connect: state?.status === 'error' ? toErrorInfo(state.error) : null,
```
raducristianpopa, line 95: Just for my information, `state` represent the transient state?

src/popup/components/ConnectWalletForm.tsx
```diff
@@ -153,23 +159,19 @@ export const ConnectWalletForm = ({
...
   );
 
-  const handleSubmit = async (ev: React.FormEvent<HTMLFormElement>) => {
-    ev.preventDefault();
-    if (!walletAddressInfo) {
-      setErrors((_) => ({ ..._, walletAddressUrl: 'Not fetched yet?!' }));
-      return;
-    }
+  const handleSubmit = async (ev?: React.FormEvent<HTMLFormElement>) => {
+    ev?.preventDefault();
```
raducristianpopa, line 166: Why is the event optional now?

src/popup/components/ConnectWalletForm.tsx
```diff
@@ -153,23 +159,19 @@ export const ConnectWalletForm = ({
...
 
     const errWalletAddressUrl = validateWalletAddressUrl(walletAddressUrl);
     const errAmount = validateAmount(amount, currencySymbol.symbol);
     if (errAmount || errWalletAddressUrl) {
       setErrors((_) => ({
         ..._,
-        walletAddressUrl: errWalletAddressUrl ? t(errWalletAddressUrl) : '',
-        amount: errAmount && t(errAmount),
+        walletAddressUrl: toErrorInfo(errWalletAddressUrl),
+        amount: toErrorInfo(errAmount),
```
raducristianpopa, line 174: We are mixing errors as exceptions and errors as value with the `validateWalletAddressUrl` and `validateAmount`. Not a big fan, but I understand why this approach.

It might be worth looking into [neverthrow](https://github.com/supermacro/neverthrow) later on.

src/popup/components/ConnectWalletForm.tsx
sidvishnoi: Because now we manually trigger the submit (without the form also), like on "Accept consent" or "retry".

src/popup/components/ConnectWalletForm.tsx
sidvishnoi: If we allow passing `{ message: string }` (which matches `Error`), we've to always add a check for `isErrorWithKey`, which is more expensive then `typeof err === 'string'`.

I'll look into rationalizing error types as a follow-up. I basically want only `Error` (`{ messsage: string }`) and `ErrorWithKeyLike`, with `failure` or `ErrorResponse` allowing returning serialization of either.

src/popup/components/ConnectWalletForm.tsx
sidvishnoi: I realize it's not used in this PR, but only in https://github.com/interledger/web-monetization-extension/pull/637. Will remove for this PR.

raducristianpopa (approved):

PR #644, opened 2024-10-05
feat: add idle detection

### LINKED ISSUE #454: Add idle detection
### Context

Post BETA

### Todos

_TBD_

raducristianpopa:
## Context

Fixes https://github.com/interledger/web-monetization-extension/issues/454.

### DIFF
--- a/src/content/container.ts
+++ b/src/content/container.ts
@@ -9,6 +9,7 @@ import {
   type ContentToBackgroundMessage,
   MessageManager,
 } from '@/shared/messages';
+import { IdleDetection } from './services/idleDetection';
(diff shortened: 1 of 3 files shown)


tomayac (commented):
For your interest, Chromium supports an [Idle Detection API](https://developer.mozilla.org/en-US/docs/Web/API/Idle_Detection_API) natively, though it's gated behind a permission request. 

In Chrome extensions, you have access to a [similar API](https://developer.chrome.com/docs/extensions/reference/api/idle), which might be of interest as well. It likewise requires permission at the extension level.

PR #647, opened 2024-10-07
test(e2e): use labels with i18n, don't hardcode

sidvishnoi:
Labels/text changed and we forgot to update E2E tests. Maybe we should run these on each PR also?


sidvishnoi (commented):
test-e2e

raducristianpopa (approved):
Let's keep them on a manual basis for now or specify when NOT to run the e2e tests (ie. renovate PRs)?

PR #656, opened 2024-10-11
refactor(popup): move ConnectWallet route out of Settings

sidvishnoi:
Preparing to make Settings route show the new tab based UI.

### DIFF
--- a/src/popup/pages/ConnectWallet.tsx
+++ b/src/popup/pages/ConnectWallet.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import { ConnectWalletForm } from '@/popup/components/ConnectWalletForm';
+import { useMessage, usePopupState } from '@/popup/lib/context';
+import { getWalletInformation } from '@/shared/helpers';
+
+export const Component = () => {
+  const { state } = usePopupState();
+  const message = useMessage();
+
+  const connectState = state.transientState['connect'];
+  return (
+    <ConnectWalletForm
+      publicKey={state.publicKey}
+      state={connectState}
+      defaultValues={{
+        recurring:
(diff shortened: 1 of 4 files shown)


src/popup/pages/Settings.tsx
```diff
@@ -1,43 +1,9 @@
...
-          // The popup closes due to redirects on connect, so we don't need to
-          // update any state manually.
-          // But we reload it, as it's open all-time when running E2E tests
-          window.location.reload();
-        }}
-        clearConnectState={() => message.send('CONNECT_WALLET', null)}
-      />
-    );
-  }
+  return <WalletInformation info={state} />;
```
raducristianpopa, line 8: Should we move the implementation for `WalletInformation` here in `pages/Settings.tsx` directly? It feels a bit redundant if we only return a component.

src/popup/pages/ConnectWallet.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { useMessage, usePopupState } from '@/popup/lib/context';
+import { getWalletInformation } from '@/shared/helpers';
+
+export const Component = () => {
+  const { state } = usePopupState();
+  const message = useMessage();
+
+  const connectState = state.transientState['connect'];
+  return (
+    <ConnectWalletForm
```
raducristianpopa, line 12: Same for `ConnectWalletForm`?

src/popup/pages/Settings.tsx
sidvishnoi: I'm going this way after this PR: https://github.com/interledger/web-monetization-extension/pull/657, in case this is what you're saying.

raducristianpopa (approved):

Issue #568, opened 2024-09-02
The payment rate should be a global setting


RabebOthmani:
### Context

This has come up as a feedback from both the community and team members. The payment rate is not per page but currently is on the payment/website page. This might cause confusion for users. It's better to move it to settings. 

### Todos

- Design the setting page or pages for that matter considering the exception list coming in the future.
- Prioritize when is the right time for this work 
- Detail technical to-do

RabebOthmani:
Meeting notes: rate of pay to move into settings 
                         unable/disable continuous payment is also a global setting --> moves to settings
                         Design settings page to tabs (wallet, rate) 
                         Add nice graphic to  pay page so it's not empty 
                         Bring back a way to globally disable/enable WM

RabebOthmani:
### Acceptance Criteria
- The settings page will have 2 tabs: Wallet address and budget.
- Under the wallet tab, the user will have the ability to see the key used to connect to their wallet, copy or generate it if needed. 
- Under the wallet tab, the user will have the ability to see the wallet/ payment pointer the extension is connected to. 
- Under the wallet tab, the user will have the ability to disconnect their wallet. 
- Under the budget tab, the user will have the ability to see: their budget, the remaining balance and the current rate of pay. 
- If the budget is set to renew monthly, the toggle will be enabled and the user will see the renewal date. 
- If the budget is not set to renew monthly, the toggle will be disabled. 
- the user have the ability to enable/disable continuous payment.
- the user have the ability to adjust the rate of pay. 
![budget setting ](https://github.com/user-attachments/assets/4156362a-31c8-4a29-9541-9463a9826c73)

PR #657, opened 2024-10-11
feat(popup/Settings)!: split settings into tabbed interface

sidvishnoi:
## Context

- Closes https://github.com/interledger/web-monetization-extension/issues/568


sidvishnoi:
## Screenshots

<details>
<summary>Wallet screen</summary>

</details>

<details>
<summary>Wallet screen, advanced expanded</summary>

</details>

<details>
<summary>Budget screen</summary>

</details>

<details>
<summary>Rate of pay screen</summary>

</details>

<details>
<summary>Rate of pay screen, WM disabled</summary>

</details>

sidvishnoi:
> For the rate of pay, I thought we were getting rid of the slider?

Eventually yes, but not in this PR. It'll get too big, and that input isn't as simple if done right.

> What happens when the user changes the rate of pay or budget amount?

Rate of pay gets handled immediately, like before.
Budget can't be changed at the moment. It's a big change. Those fields are disabled/readonly.

RabebOthmani:
Ok for rate of pay slider. 
As for the budget, we can't just say something is not going to be done because it's big change. The requirement was never for it to be read-only. 
In the last call, we agreed that they can change the budget but it will completely override the previous grant and starts immediately. @raducristianpopa keep me honest here

sidvishnoi:
We'd need UI to support edits first then. We need submit buttons somewhere, and a messaging that they'll need to connect wallet again.

RabebOthmani:
Why do they need to connect to the wallet again?

RabebOthmani:
Does that mean that switching between monthly or not will cause the same issues

sidvishnoi:
When they change budget (any part of it - amount or recurring), we need to create a new interactive grant (i.e. ask user for permission), which happens in a new tab. So, it's the same connect wallet process again, but only without the need to add key again.

RabebOthmani:
You mean the part on the browser to accept the grant? or the connect screen on the extension? 
That also mean that the monthly toggle is the wrong control. A toggle is to make changes not just for visual effects

sidvishnoi:
> You mean the part on the browser to accept the grant?

Yes.

>  That also mean that the monthly toggle is the wrong control. A toggle is to make changes not just for visual effects

Yes, wrong control if it's not meant to be editable. [Point #3 in Slack](https://interledgerfoundation.slack.com/archives/C06CVJRT5N3/p1728391732857209?thread_ts=1728391192.246379&cid=C06CVJRT5N3)

RabebOthmani:
Ok. Settings should be adjustable. A user should have the ability to change the budget allocated to use with the extension and whether or not they want to renew it monthly. 
They will have to accept the grant online. That is absolutely fine. 
We will add a submit changes button that gets enabled if the amount or renew values change. Similar to connect, when they click , they get redirected to the browser. 

Please do let me know if anything is not clear or if there are any technical limitations before deciding on requirements on your own. 
budget

sidvishnoi:
Will do budget-edit support as a follow-up PR. Keeping this PR limited to tab-UI thing only.

src/popup/components/Settings/WalletInformation.tsx
```diff
@@ -0,0 +1,70 @@
...
+  } = useForm();
+
+  return (
+    <div className="flex h-full flex-col gap-8">
+      {/* TODO: Improve error handling */}
+      <form
+        className="space-y-4"
+        onSubmit={handleSubmit(async () => {
+          await message.send('DISCONNECT_WALLET');
+          window.location.reload();
```
raducristianpopa, line 33: We should check the response here. But I guess it's going to be part of the above TODO item, right?

raducristianpopa (approved):

Issue #674, opened 2024-10-16
Make budget editable


sidvishnoi:
Follow-up of https://github.com/interledger/web-monetization-extension/pull/657. See https://github.com/interledger/web-monetization-extension/pull/657#issuecomment-2414402814 for design reference.

PR #675, opened 2024-10-16
feat: support modifying budget

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/674

- Add `UPDATE_BUDGET` handler
- Add submit button on budget screen, which is enabled only if amount/recurring is changed from existing state.
- Could not use `addFunds` logic as:
  - we need to revoke all existing grants, not just the ones of same type like addFunds does
  - we also need to clear other grant from storage
  - different intent (for redirect screen)


src/background/services/openPayments.ts
```diff
@@ -459,6 +464,21 @@ export class OpenPaymentsService {
...
+      'oneTimeGrant',
+      'recurringGrant',
+    ]);
+
+    await this.completeGrant(
+      amount,
+      walletAddress!,
+      recurring,
+      InteractionIntent.BUDGET_UPDATE,
+    );
```
sidvishnoi, line 479: TODO:
- complete new grant
- on success, revoke existing grants (both recurring and one-time)
- on failure, keep existing grants
- reflect progress/errors into popup

RabebOthmani:
The change to my budget was never reflected on the wallet but then I made a OTP and I could see it

sidvishnoi:
Discussed above, the new budget gets reflected in grants page, not on transactions. So, all good.

Issue #612, opened 2024-09-19
[BUG] sending a one-time payment reports success even when there is no money on the account


asurkov:
### Steps to reproduce

1. Go to https://ericahargreave.com/
2. Click on extension button
3. Set 1.00 for the amount
4. Click send

### Expected result

The failure should be reported

### Actual result

The success is reported

### Screenshots or videos

https://github.com/user-attachments/assets/69d11cdb-e61b-4a5f-96ae-2030b5817498

### Operating system

macOS

### Browsers

Firefox

### Extension version

0.1.1

sidvishnoi:
Duplicate of #592
Duplicate of #567

It's on our radar. Closing this to keep tracking better.

sidvishnoi:
I'll keep this open so we can close 3 issues at once lol.

Issue #592, opened 2024-09-12
[BUG] Misleading successful payment msg even when wallet payment fails


tselit:
### Steps to reproduce

**Tester feedback**
Used `rafiki.money` to setup 2 different wallets, to easily track my WM sending wallet (starting balance of $50) and receiving wallet (starting balance of $0). 
Both wallets are in USD, they're configured to have WM-enabled payment pointers.

- connect WM extension to the sending wallet.
- setup the WM playground to receive payments into 2 payment pointers on the receiving wallet
- use the WM extension to make continuous payments to the WM playground
- use the WM extension to make a one-time payment of $2.50, which split the amount in half, to send to each of the 2 receiving pay-pointers

**Observation**: The extension shows a "Payment successful" message
**Results**: Waited 5 minutes...30 minutes...about 20 hrs but there were no successful, or even rejected transactions, and the balances remained unchanged, for each wallet.

### Expected result

None of the payments from the WM extension actually succeeded, and so the extension should **not** show a message indicating successful payment.

### Actual result

When making a one-time payment, the extension showed a "Payment successful" message even though the transaction didn't actually succeed.

### Additional context

I was concerned that maybe the WM playground or extension may be doing something unexpected.
When I repeat the scenario using a Fynbos wallet, the payments go through.
Erica didn't use the WM playground, instead she added the receiving payment pointers onto a website, but unfortunately she made similar observations.

### Operating system

Windows

### Operating system version

10

### Browsers

Chrome

### Browser version

Version 128.0.6613.115 (Official Build) (64-bit)

### Extension version

private beta release

sidvishnoi:
@tselit Can you share the wallet address of:
- connected wallet
- and the website?

tselit:
Sure @sidvishnoi . The website was the WM playground, and the connected wallet was: 
$ilp.rafiki.money/3rd-new-wm  
(this was created specifically for WM, with the "I want to use this payment pointer for Web Monetization" checkbox ticked)

sidvishnoi:
So me and Radu couldn't both reproduce with non-WM payment pointers.
Also note, WM payment pointers are going away soon. The regular ones will work for WM.

Issue #569, opened 2024-09-02
[UX Improvement] Payment page 


RabebOthmani:
### Context

Once the extension is all setup to facilitate payments, few things on the payment page can be improved for a better experience for the users. 
For example: the way continuous payments and one off payments are presented may confuse the users into thinking it is one flow  

### Todos

- Revamp the page layout  - Check out Figma for design 
  - https://github.com/interledger/web-monetization-extension/issues/568
  -  https://github.com/interledger/web-monetization-extension/issues/567
  - Update message 'WM has been turned off' to 'ongoing payments are disabled' 
- Review copy and make it easier for users to understand ( example: continuous payment stream may not be easy to understand)

Issue #567, opened 2024-09-02
[FEATURE REQUEST] Improve payment confirmation 


RabebOthmani:
### Describe the feature you would like to request

**_This request is based on tester's feedback_** 

After sending a payment, the confirmation within the button happens quickly and then resets. There's a risk of the user not noticing the confirmation if they got slightly distracted away from that screen. 

### Describe the solution you would like

Stay in confirmed state until user takes an action (maybe an ok or return button ) 

### Acceptance criteria

- The extension should show an error message if not able to facilitate the payment with the text ' **Something went wrong. Please try again'**
- The extension should show a confirmation message if able to facilitate the payment with the text **' Thanks for your support.'**
- Once a payment is facilitated, the amount is reset to 0 and the 'pay now' button is disabled. 
- The text above the amount should say **' Support [ website url]'**
- the 'pay now' button is enabled once the amount is > 0. 
~~- The extension should say 'donate now' instead of 'pay now'~~ 

**_Screenshot is for prototyping purposes only and not a final design_**
<img src="https://github.com/user-attachments/assets/121b978b-1c96-48a9-a8cb-5e2984949784" width="300">

melissahenderson:
Suggestion: instead of "Thank you for supporting . You can now make more donations."
How about: "Thanks for your support." It sounds natural and stays away from the other terms we're chatting about in Slack.
Then the action button can be "Close"

As far as "Pay Now" and etc, it looks like those conversations are still in progress.

RabebOthmani:
@sidvishnoi @raducristianpopa the design for the payment page and different OTP states is available on Figma.

PR #694, opened 2024-11-01
fix(background)!: better payment confirmation & message

sidvishnoi:
## Context

- Closes https://github.com/interledger/web-monetization-extension/issues/567
- Closes https://github.com/interledger/web-monetization-extension/issues/592
- Closes https://github.com/interledger/web-monetization-extension/issues/569
- Closes https://github.com/interledger/web-monetization-extension/issues/445
- Closes https://github.com/interledger/web-monetization-extension/issues/612

- Poll for outgoing payment.
  - We poll every 1.5s for up to total 8s, with initial delay also of 1.5s.
  - Polling success is defined as `sentAmount == debitAmount`.
  - We keep emitting latest `OutgoingPayment` state, so even if polling isn't success, we may have partial payment.
  - Error state corresponds to: `OutgoingPayment.failed = true` and `sentAmount == 0`
  - Payment warning corresponds to: polling timeout reached and sentAmount is still zero.
  - We start polling immediately, as part of `pay({ amount })`
  - I'm not sure if it'll be better to inform user of these two parts, as polling can show the spinner for up to additional 8s. We might want to show user two parts to avoid showing infinite spinner: "Sending..." ... "Verifying payment"?
  - **Added `read` permission on outgoing-payment** (breaking change as calls from existing connected extensions will throw on read), but avoided breaking change by catching the error and checking `debitAmount` from initial outgoing payment response like before.
- Update UI to show error/warning/success message; remove "Payment successful" button state.


src/background/services/openPayments.ts
```diff
@@ -694,7 +695,7 @@ export class OpenPaymentsService {
             },
             {
               type: 'outgoing-payment',
-              actions: ['create'],
+              actions: ['create', 'read'],
```
sidvishnoi, line 698: This is the breaking change, as already users who have already connected extension will not have this permission in their tokens.
Need to handle this in `outgoingPayment.get()` by checking for "Inactive token" error, to make this a non-breaking change.

src/background/services/openPayments.ts
```diff
@@ -881,6 +882,43 @@ export class OpenPaymentsService {
...
+    outgoingPaymentId: OutgoingPayment['id'],
+    {
+      signal,
+      maxAttempts = 10,
+    }: Partial<{ signal: AbortSignal; maxAttempts: number }> = {},
+  ) {
+    let attempt = 0;
+    let outgoingPayment: undefined | OutgoingPayment;
+    while (++attempt <= maxAttempts) {
+      signal?.throwIfAborted();
```
sidvishnoi, line 895: Supported well enough for our use 🎉 
https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted#browser_compatibility

src/background/services/openPayments.ts
```diff
@@ -881,6 +882,43 @@ export class OpenPaymentsService {
     return outgoingPayment;
   }
 
+  async outgoingPaymentWaitForCompletion(
```
sidvishnoi, line 885: Maybe shorten the name to `pollOutgoingPayment`?

src/background/services/monetization.ts
```diff
@@ -269,18 +276,61 @@ export class MonetizationService {
...
+            { signal, maxAttempts: OUTGOING_PAYMENT_POLLING_MAX_ATTEMPTS },
+          )) {
+            outgoingPayments.set(sessionId, outgoingPayment);
+          }
+          return outgoingPayments.get(sessionId)!;
+        }),
+    );
+
+    const totalSentAmount = [...outgoingPayments.values()].reduce(
+      (acc, op) => acc + BigInt(op?.sentAmount?.value ?? 0),
```
sidvishnoi, line 302: We earlier had `debitAmount` here.

src/background/services/openPayments.ts
```diff
@@ -881,6 +884,50 @@ export class OpenPaymentsService {
...
+          await this.rotateToken();
+        } else {
+          throw error;
+        }
+      }
+    }
+
+    throw new ErrorWithKey(
+      'pay_warn_pay_warn_outgoingPaymentPollingIncomplete',
+    );
```
raducristianpopa, line 928:
```suggestion
    throw new ErrorWithKey(
      'pay_warn_outgoingPaymentPollingIncomplete',
    );
```

src/background/services/openPayments.ts
```diff
@@ -1023,6 +1070,15 @@ export const isOutOfBalanceError = (error: any) => {
   return error.status === 403 && error.description === 'unauthorized';
 };
 
+export const isMissingGrantPermissionsError = (error: any) => {
+  if (!isOpenPaymentsClientError(error)) return false;
+  return (
+    error.status === 403 &&
+    (error.description === 'Insufficient Grant' /* Fynbos */ ||
+      error.description === 'Inactive Token')
```
raducristianpopa, line 1078: We shouldn't check for `Inactive Token` here. We are checking for this description with `isTokenExpiredError`
```suggestion
      error.description === 'Inactive Token')
```

src/popup/components/PayWebsiteForm.tsx
```diff
@@ -42,54 +34,72 @@ export const PayWebsiteForm = () => {
...
-      setErrors((prev) => ({ ...prev, pay: toErrorInfo(response.message) }));
+      setErrors((prev) => ({
+        ...prev,
+        pay: toErrorInfo(response.error || response.message),
+      }));
     } else {
-      setButtonState('success');
       setAmount('');
+      const { type, url, sentAmountFormatted } = response.payload;
+      const msg = t('pay_state_success', [sentAmountFormatted, url]);
```
raducristianpopa, line 60: I feel like adding website URL again in the message is not really needed since the label explicitly says: `Pay {website}`.

src/background/services/monetization.ts
```diff
@@ -263,27 +280,95 @@ export class MonetizationService {
...
+      ]),
+    );
+    this.logger.debug('polling outgoing payments for completion');
+    const signal = AbortSignal.timeout(OUTGOING_PAYMENT_POLLING_MAX_DURATION); // can use other signals as well, such as popup closed etc.
+    const pollingResults = await Promise.allSettled(
+      [...outgoingPayments]
+        .filter(([, outgoingPayment]) => outgoingPayment !== null)
+        .map(async ([sessionId, outgoingPaymentInitial]) => {
+          for await (const outgoingPayment of this.openPaymentsService.pollOutgoingPayment(
+            outgoingPaymentInitial!.id,
```
raducristianpopa, line 305: Maybe add a comment that points to this issue? 
https://github.com/microsoft/TypeScript/issues/41173

src/background/services/monetization.ts
```diff
@@ -263,27 +280,95 @@ export class MonetizationService {
...
+        };
+      }
+
       const isNotEnoughFunds = results
         .filter((e) => e.status === 'rejected')
         .some((e) => isOutOfBalanceError(e.reason));
+      const isPollingLimitReached = pollingErrors.some(
+        (err) =>
+          (isErrorWithKey(err) &&
+            err.key === 'pay_warn_pay_warn_outgoingPaymentPollingIncomplete') ||
```
raducristianpopa, line 347:
```suggestion
            err.key === 'pay_warn_outgoingPaymentPollingIncomplete') ||
```

src/background/services/openPayments.ts
```diff
@@ -1023,6 +1070,15 @@ export const isOutOfBalanceError = (error: any) => {
   return error.status === 403 && error.description === 'unauthorized';
 };
 
+export const isMissingGrantPermissionsError = (error: any) => {
+  if (!isOpenPaymentsClientError(error)) return false;
+  return (
+    error.status === 403 &&
+    (error.description === 'Insufficient Grant' /* Fynbos */ ||
```
raducristianpopa, line 1077: What does the `Fynbos` comment represent?

src/popup/components/PayWebsiteForm.tsx
```diff
@@ -42,54 +34,72 @@ export const PayWebsiteForm = () => {
 
   const form = React.useRef<HTMLFormElement>(null);
   const [isSubmitting, setIsSubmitting] = React.useState(false);
-  const [buttonState, setButtonState] =
-    React.useState<keyof typeof BUTTON_STATE>('idle');
-  const isIdle = React.useMemo(() => buttonState === 'idle', [buttonState]);
+  const [msg, setMsg] = React.useState<null | {
+    type: 'success' | 'warn';
+    message: string;
+  }>(null);
```
raducristianpopa, line 40: Maybe this is going to be more explicit?
```suggestion
  const [payStatus, setPayStatus] = React.useState<null | {
    type: 'full' | 'partial';
    message: string;
  }>(null);
```

src/_locales/en/messages.json
```diff
@@ -48,16 +48,37 @@
...
+    "message": "Send now"
+  },
   "pay_error_notEnoughFunds": {
     "message": "Insufficient funds to complete the payment."
   },
+  "pay_error_outgoingPaymentFailed": {
+    "message": "We were unable to process this transaction. Please try again!",
+    "description": "OutgoingPayment.failed === true"
+  },
+  "pay_warn_pay_warn_outgoingPaymentPollingIncomplete": {
```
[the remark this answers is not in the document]
sidvishnoi: Good catch!

src/background/services/openPayments.ts
sidvishnoi: Fynbos shows this error as "Insufficient grant", while test wallet fails with "Inactive token". Quite weird I know, but this check passes both.

src/popup/components/PayWebsiteForm.tsx
sidvishnoi: ➕

src/_locales/en/messages.json
```diff
@@ -48,16 +48,37 @@
...
   },
+  "pay_action_pay": {
+    "message": "Send now"
+  },
   "pay_error_notEnoughFunds": {
     "message": "Insufficient funds to complete the payment."
   },
+  "pay_error_outgoingPaymentFailed": {
+    "message": "We were unable to process this transaction. Please try again!",
+    "description": "OutgoingPayment.failed === true"
```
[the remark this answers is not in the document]
sidvishnoi: Just to be clear, `OutgoingPayment.failed = true` and `OutgoingPayment.sent_amount > 0` is possible?

We want to treat any partial amount to be treated as success here?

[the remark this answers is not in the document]
raducristianpopa: > Just to be clear, OutgoingPayment.failed = true and OutgoingPayment.sent_amount > 0 is possible?
Yes.

> We want to treat any partial amount to be treated as success here?
We cannot treat a sent partial amount as an error (`failed = true && sent_amount = 0`), since an amount of was actually sent. We can use `payStatus.type = 'partial'` for this.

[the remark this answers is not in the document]
sidvishnoi: Discussed cases with @raducristianpopa on Slack. We need to handle these cases:
```
failed = T; sent_amount = 0             => definitely error. immediate fail.
failed = F; sent_amount = 0             => we don't know yet. maybe warn user? It could be an error if check for long enough, or sent_amount might become non-zero

failed = T; sent_amount > 0             => ok, but only a partial was sent
failed = F; sent_amount > 0             => ok, but only a partial was sent, can be successfully completed afterwards

failed = T; sent_amount = debit_amount  => not possible
failed = F; sent_amount = debit_amount  => ok. immediate success.
```

@RabebOthmani How to best show this to user? For added context, after the payment, we've to repeatedly check payment status (polling). Right now, we check for up to 8 seconds only. The payment might be in any of above 5 states.

Also, note that, we'll be showing user a spinner for those 8s. Might be better to show "verifying payment status" message while we are in verify stage?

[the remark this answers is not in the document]
melissahenderson: 8s is a long time for just a spinner. My unsolicited opinion is that we should show some kind of messaging for sure.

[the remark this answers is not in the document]
RabebOthmani: Per Slack: 
* " It looks like this is taking long. Check your wallet later to ensure the payment went through" 
* Partial payment- only <amount> was successfully sent. to learn more <link>

src/background/services/openPayments.ts
sidvishnoi: Resolution: since Rafiki alpha.16, we get "Inactive Token" only.

src/background/services/openPayments.ts
```diff
@@ -881,6 +884,59 @@ export class OpenPaymentsService {
...
+          if (isTokenInactiveError(error) && !tokenRotated) {
+            // isMissingGrantPermissionError has same error msg as isTokenInactiveError
+            tokenRotated = true;
+            await this.rotateToken();
+            continue;
+          }
+          throw error;
+        } else if (isTokenExpiredError(error) && !tokenRotated) {
+          tokenRotated = true;
+          await this.rotateToken();
```
sidvishnoi, line 930: Maybe a better idea would be check response of `rotateToken` to see if token as required permission. Will do.

sidvishnoi:
> We do not have a message for a partial payment?

Yep, we treat it as regular payment, and as success (`pay_state_success` key). Not sure if showing partial amount in message is a good idea though - might confuse users.

RabebOthmani:
@sidvishnoi all good on here ? Do you need more input from me?

sidvishnoi:
@RabebOthmani Two points actually (can be addressed as follow-up I guess, as this PR is pending since too long)

(1)

> Also, note that, we'll be showing user a spinner for those 8s. Might be better to show "verifying payment status" message while we are in verify stage? https://github.com/interledger/web-monetization-extension/pull/694#discussion_r1836833050

> 8s is a long time for just a spinner. My unsolicited opinion is that we should show some kind of messaging for sure. https://github.com/interledger/web-monetization-extension/pull/694#discussion_r1840242671


(2)
> > We do not have a message for a partial payment? https://github.com/interledger/web-monetization-extension/pull/694#pullrequestreview-2458656675

> Yep, we treat it as regular payment, and as success (`pay_state_success` key). Not sure if showing partial amount in message is a good idea though - might confuse users. https://github.com/interledger/web-monetization-extension/pull/694#issuecomment-2498250104

raducristianpopa (approved):

PR #712, opened 2024-11-13
fix(scripts/build): use correct asset-copy paths on Windows 

### LINKED ISSUE #711: [BUG] Fix extension build windows environment
### Steps to reproduce

1. Run the esbuild build process on a Windows machine.
2. Observe that the `copy` plugin fails to copy the files correctly in either production or development build dirs

### Expected result

The copy plugin should be able to copy all the files correctly, regardless of the operating system (Windows, Linux, or macOS).

### Actual result

When using the copy plugin from the esbuild-plugin-copy package on Windows, the file paths are not being correctly interpreted, leading to the plugin failing to copy the files.

Different file path conventions between Windows and other operating systems like Linux and macOS. On Windows, file paths typically use backslashes ( \ ) as the directory separator, while on Linux and macOS, the standard directory separator is a forward slash ( / ).

The copy plugin from the `esbuild-plugin-copy` package is designed to work with file paths using forward slashes, as this is the standard convention for file paths in js and most modern web development tools. When you use backslashes on Windows, the copy plugin is unable to properly interpret the file paths, leading to this issue.

### Screenshots or videos

image

- missing `_locales`
- missing `assets`
- missing `pages/progress-connect/index.html`
- missing `popup/index.html`

### Additional context

To resolve this problem on Windows, we need to convert the backslashes to forward slashes in the `from` and `to` paths passed to the copy plugin.

### Operating system

Windows, Linux

### Operating system version

_No response_

### Browsers

Firefox

### Browser version

_No response_

### Extension version

0.1.0

DarianM:
## Context

The copy plugin should be able to copy all the files correctly, regardless of the operating system (Windows, Linux, or macOS).

Closes #711 .

Convert the backslashes to forward slashes in the from and to paths passed to the copy plugin.

### DIFF
--- a/esbuild/plugins.ts
+++ b/esbuild/plugins.ts
@@ -80,26 +80,26 @@ function getAssetPaths(srcDir: string, outDir: string) {
   const assetPaths = [
     {
       from: path.join(srcDir, 'popup', 'index.html'),
-      to: path.join(outDir, 'popup', 'index.html')
+      to: path.join(outDir, 'popup', 'index.html'),
     },
     {
       from: path.join(srcDir, 'pages', 'progress-connect', 'index.html'),
-      to: path.join(outDir, 'pages', 'progress-connect', 'index.html')
+      to: path.join(outDir, 'pages', 'progress-connect', 'index.html'),
     },
     {
       from: path.join(srcDir, '_locales', '**', '*'),
-      to: path.join(outDir, '_locales')
+      to: path.join(outDir, '_locales'),
     },
     {
       from: path.join(srcDir, 'assets', '**', '*'),
-      to: path.join(outDir, 'assets')
-    }
+      to: path.join(outDir, 'assets'),
+    },
   ];
 
   // normalize paths to use forward slashes
   return assetPaths.map(({ from, to }) => ({
     from: from.split(sep).join('/'),
-    to: to.split(sep).join('/')
+    to: to.split(sep).join('/'),
   }));
 }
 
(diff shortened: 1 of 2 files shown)


sidvishnoi (commented):
I wonder if we can directly use `path.posix.join()` instead to force forward slashes everywhere.

DarianM:
@sidvishnoi 
we can, but the issue is that `path.posix.join()` assumes all input paths are using forward slashes, but the input paths (`SRC_DIR`, `DEV_DIR`, `DIST_DIR` ) are actually provided by the operating system and nodejs, which means on Windows they will contain backslashes.

```
// if we use path.posix.join():
path.posix.join(SRC_DIR, 'popup', 'index.html')
// results in: 'D:\\web-monetization-extension\\src/popup/index.html'
//  it doesn't convert the existing backslashes in the input paths
```

we can use path.posix directly, would first need to convert all input paths to use forward slashes from `esbuild/config.ts` like 
`export const DEV_DIR = path.resolve(ROOT_DIR, 'dev').split(sep).join('/');`

sidvishnoi (commented):
I think cleanest would be adding a util (say `toPosix`) and just call it to replace all windows separators to posix type, only for the copy plugin.

esbuild/plugins.ts
```diff
@@ -1,4 +1,4 @@
-import path from 'node:path';
+import path, { posix, sep } from 'node:path';
```
sidvishnoi, line 1: Nit: let's not import these separately and continue using `path.` namespace.

sidvishnoi (approved):
Perfect. Thanks @DarianM and congrats on your first PR here 🎉

PR #714, opened 2024-11-17
feat(ErrorKeyRevoked): reconnect with automatic key addition 

### LINKED ISSUE #653: Use automatic key addition when key revoked during use
With https://github.com/interledger/web-monetization-extension/issues/615, we can now automatically add keys to wallet. We should extend its use on key revoked state.

In `src/popup/pages/ErrorKeyRevoked.tsx`, use `KeyAutoAdd` to add keys, instead of directly asking user to manually paste key.

DarianM:
## Context

When the public key is revoked, instead of manually asking to copy/pasta the key to reconnect the wallet, use `KeyAutoAddService` to automatically add the key back,

Closes #653 .

~EDIT: depends on #742 to be done~

- add publicKey to wallet automatically (with user's consent) then redirect to welcome page


src/background/services/openPayments.ts
```diff
@@ -112,13 +112,15 @@ const enum ErrorCode {
   KEY_ADD_FAILED = 'key_add_failed',
 }
 
-const enum GrantResult {
-  SUCCESS = 'grant_success',
-  ERROR = 'grant_error',
+const enum Result {
```
raducristianpopa, line 115: Lets revert the name for the enum.

src/background/services/openPayments.ts
```diff
@@ -112,13 +112,15 @@ const enum ErrorCode {
...
+  GRANT_SUCCESS = 'grant_success',
+  GRANT_ERROR = 'grant_error',
+  KEY_ADDED = 'key_added',
 }
 
 const enum InteractionIntent {
```
raducristianpopa, line 121: @sidvishnoi can we get a list with the missing intents/results for the welcome page? IIRC we discussed some of them before the summit, but I cannot recall which ones.

src/background/services/openPayments.ts
```diff
@@ -906,7 +908,16 @@ export class OpenPaymentsService {
 
   async reconnectWallet() {
     try {
+      const { walletAddress } = await this.storage.get(['walletAddress']);
+      const tabId = await this.addPublicKeyToWallet(walletAddress!);
```
raducristianpopa, line 912: Lets do a proper check for the wallet address instead of performing a non-null assertion.

src/background/services/openPayments.ts
```diff
@@ -906,7 +908,16 @@ export class OpenPaymentsService {
...
+      this.setConnectState('connecting');
       await this.rotateToken();
+      await this.redirectToWelcomeScreen(
+        tabId!,
```
raducristianpopa, line 916: The tab should be open when we redirect to the welcome screen, otherwise the redirect will throw.

CC @sidvishnoi: should we still perform the `tabId` check as well? If the user closes the tab we can open a new one if needed.

src/popup/components/ErrorKeyRevoked.tsx
```diff
@@ -164,16 +164,15 @@ const ReconnectScreen = ({
...
+      <div className="space-y-2 text-sm">
         <p className="px-2">
           Reconnecting to wallet:{' '}
           <span className="underline">{info.walletAddress?.id}</span>
         </p>
         <p className="px-2">
-          <strong>Before</strong> you reconnect, copy the public key below and
-          add it to your wallet.
+          <strong>Reconnect</strong> to automatically add the key to your
+          wallet.
```
raducristianpopa, line 174: We should discuss this in the call: giving the users option to choose how to add the key - automatically or manually.

sidvishnoi: Yes, we'll do that, reusing same consent screen. But keeping PR smaller atm.

sidvishnoi: But still good to discuss in call also.

src/background/services/openPayments.ts
sidvishnoi: results:
- `key_add_failed` or `key_add_error`
- `key_added` or `key_add_success` (this one)

intents:
- connect
- add_funds
- reconnect (this one)
- update_budget (missing from earlier)

src/background/services/openPayments.ts
sidvishnoi: Might be better to update `Storage` type? When wallet is in `connected` state, certain properties in `Storage` will not be nullish. 
Or maybe better to have runtime checks so user messing with storage (or something else going wrong) still gives us better errors?

src/popup/components/ErrorKeyRevoked.tsx
```diff
@@ -50,6 +49,7 @@ export const ErrorKeyRevoked = ({
           onReconnect={() => {
             clearScreen();
             onReconnect?.();
+            window.close();
```
sidvishnoi, line 52:
```suggestion
            window.reload();
```

src/popup/components/ErrorKeyRevoked.tsx
sidvishnoi: In either case, we should not be showing this screen.
The auto-key-add-reconnect should have been triggered from `MainScreen` directly; or lead to consent screen followed by previous manual-key add screen.

src/popup/components/ErrorKeyRevoked.tsx
```diff
@@ -109,7 +126,11 @@ const MainScreen = ({
         <Button onClick={() => requestDisconnect()} loading={loading}>
           {t('keyRevoked_action_disconnect')}
         </Button>
-        <Button onClick={() => requestReconnect()}>
+        <Button
+          onClick={() => {
+            setScreen('consent-reconnect');
+          }}
```
sidvishnoi, line 132: Nit: inline like this:
```suggestion
          onClick={() => setScreen('consent-reconnect')}
```

src/background/services/openPayments.ts
```diff
@@ -904,16 +919,51 @@ export class OpenPaymentsService {
...
+      throw error;
+    }
+    await this.storage.setState({ key_revoked: false });
+  }
+
   async reconnectWallet() {
     try {
+      const { walletAddress } = await this.storage.get(['walletAddress']);
+      if (
+        !isWalletAddress(walletAddress) ||
```
sidvishnoi, line 939: I think just checking if `walletAddress` exists should be enough.

We can perhaps add such storage validations in `background.ts` based on storage update events - to ensure storage is not manipulated maliciously.

src/background/services/openPayments.ts
```diff
@@ -904,16 +919,51 @@ export class OpenPaymentsService {
...
+        !KeyAutoAddService.supports(walletAddress)
+      ) {
+        // this.updateConnectStateError(error);
+        throw new ErrorWithKey('connectWalletKeyService_error_notImplemented');
```
sidvishnoi, line 943: IMPORTANT: The reason we have this in `connectWallet` like that (i.e. check if `autoKeyAddConsent`) can be found in https://github.com/interledger/web-monetization-extension/pull/637#discussion_r1787599150. Likely worth following same approach here.

sidvishnoi (commented):
Nearly there...

src/background/services/openPayments.ts
```diff
@@ -947,11 +996,15 @@ export class OpenPaymentsService {
...
+  async rotateToken(
+    options: { dedupeRejected: boolean } = { dedupeRejected: true },
+  ) {
     if (!this.grant) {
       throw new Error('No grant to rotate token for');
     }
-    const rotate = this.deduplicator.dedupe(this.client!.token.rotate);
+    const rotate = this.deduplicator.dedupe(this.client!.token.rotate, {
+      dedupeRejected: options.dedupeRejected,
+    });
```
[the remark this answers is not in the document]
DarianM: In the wallet reconnect flow, when the first `validateRedirect` rejects, it is cached for `5000` ms. This translates to the second `validateRedirect` attempt after adding the key, to return the rejected promise to the cache instead of trying again with a new request.

So I do not want to cache the token rotate when it comes to redirect, to I can try again to rotate the token at reconnect.

src/background/services/openPayments.ts
```diff
@@ -917,6 +924,48 @@ export class OpenPaymentsService {
...
+            throw new Error('Could not create tab');
+          }
+          await this.redirectToWelcomeScreen(
+            tabId,
+            GrantResult.KEY_ADD_SUCCESS,
+            InteractionIntent.RECONNECT,
+          );
+        } catch (error) {
+          this.updateConnectStateError(error);
+          throw error;
```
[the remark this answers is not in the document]
DarianM: We could, but having some trouble with `tabIb` to redirect to. Do you think is a good idea to 'recycle' the `tabId` from `addPublicKeyToWallet` or create a new `tabs.create` for `KEY_ADD_ERROR` ?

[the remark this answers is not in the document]
sidvishnoi: It's always a good idea to recycle (if it's still open).
Note that, we don't want to open a tab if user explicitly closed it (need to catch that error separately if it's thrown).

src/_locales/en/messages.json
```diff
@@ -168,6 +168,9 @@
   "connectWalletKeyService_text_consentP1": {
     "message": "We will automatically connect with your wallet provider."
   },
+  "reconnectWalletKeyService_text_consentP1": {
+    "message": "We will automatically reconnect with your wallet provider."
```
DarianM, line 172: todo: ask to improve message later

src/popup/components/ErrorKeyRevoked.tsx
```diff
@@ -35,16 +36,35 @@ export const ErrorKeyRevoked = ({
...
+    );
+  } else if (screen === 'consent-reconnect') {
+    return (
+      <AnimatePresence mode="sync">
+        <AutoKeyAddConsent
+          onAccept={async () => {
+            try {
+              await reconnectWallet({ autoKeyAddConsent: true });
+              onReconnect?.();
+              clearScreen();
```
sidvishnoi, line 53: If `onReconnect` reloads page, we'll never `clearScreen`

```suggestion
              clearScreen();
              onReconnect?.();
```

sidvishnoi (approved):
LGTM! 🎉💯

src/background/services/openPayments.ts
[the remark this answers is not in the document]
DarianM: @raducristianpopa thinking you're right.
This applies wherever  a rejected `rotateToken` is  followed by another `rotateToken` withing the `wait` cache clear.

`pay`, `payContinuous`, `adjustAmount` can use cached rejected promises instead of retrying to rotate.
Especially since I see that the `pay` website can call itself recursively if the token has expired.

cc @sidvishnoi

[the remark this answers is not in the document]
sidvishnoi: Sounds right to me. Should still do a local manual test.

Aside: might be good to rename it to `cacheRejections`?

raducristianpopa (approved):

src/background/services/deduplicator.ts
```diff
@@ -22,7 +23,11 @@ export class Deduplicator {
 
   dedupe<T extends AsyncFn<any>>(
     fn: T,
-    { cacheFnArgs = false, wait = 5000 }: Partial<DedupeOptions> = {},
+    {
+      cacheFnArgs = false,
+      cacheRejections = true,
```
raducristianpopa, line 28:
```suggestion
      cacheRejections = false,
```

src/background/services/deduplicator.ts
```diff
@@ -44,6 +49,10 @@ export class Deduplicator {
           return res;
         })
         .catch((err) => {
+          if (!cacheRejections) {
+            this.cache.delete(key);
+          }
```
raducristianpopa, line 54:
```suggestion
          if (cacheRejections) {
            this.cache.delete(key);
          }
```

raducristianpopa (changes requested):

src/background/services/deduplicator.ts
```diff
@@ -44,6 +49,10 @@ export class Deduplicator {
...
+
```
raducristianpopa, line 55: What do you think about this approach?

Note: `Promise.reject(err)` should replace `throw err`.
```suggestion
          if (cacheRejections) {
            this.cache.set(key, { promise: Promise.reject(err) });
          } else {
            this.cache.delete(key);
          }
          
          return Promise.reject(err);
```

CC @sidvishnoi @DarianM

DarianM: basically this would make `cacheRejections` false by default to all `dedupe` calls. i think it's right

sidvishnoi: @DarianM Can we split changes to deduplicator to a separate PR and add unit tests to ensure it works as intended?

DarianM: @sidvishnoi sure. added #742 as depends on. let's do this first pronto

DarianM:
I would like to merge [WICG/webmonetization/481](https://github.com/WICG/webmonetization/pull/482) first,
to test the entire flow one last time

raducristianpopa (approved):

src/background/services/openPayments.ts
```diff
@@ -981,6 +989,66 @@ export class OpenPaymentsService {
...
+    }
+    if (!KeyAutoAddService.supports(walletAddress)) {
+      throw new ErrorWithKey('connectWalletKeyService_error_notImplemented');
+    }
+
+    try {
+      this.setConnectState('connecting');
+      await this.validateReconnect();
+    } catch (error) {
+      if (isInvalidClientError(error?.cause)) {
```
sidvishnoi, line 1010: One nit: let's do the inverse error handling in this if, and then rest of this code won't require nesting (early return). Like:
```
if (!isInvalidClientError()) {
  setConnectState(error)
  throw error
}
// Everything else here, without else {}
```

sidvishnoi (approved):
Suggested a code style improvement, but LGTM 🎉

sidvishnoi (approved):

PR #654, opened 2024-10-10
feat(popup): move WM toggle Switch to global header

sidvishnoi:
## Context

- Closes https://github.com/interledger/web-monetization-extension/issues/623
- Closes https://github.com/interledger/web-monetization-extension/issues/471
- Part of https://github.com/interledger/web-monetization-extension/issues/569

## Changes proposed in this pull request

Move WM toggle switch to global header.
Visible only when wallet is connected, and in non-error state.

Screenshots:

<details>
<summary>Not visible in error state</summary>
Not visible in error state
</details>

<details>
<summary>Not visible when wallet not connected</summary>
Not visible when wallet not connected
</details>

<details>
<summary>Can toggle from any page</summary>
Can toggle from any page
</details>

<details>
<summary>As of this PR, slider is still shown on payment page</summary>
As of this PR, slider is still shown on payment page
</details>

<details>
<summary>This message can be improved</summary>
This message can be improved
</details>

Issue #623, opened 2024-09-24
Ability to disable WM 


RabebOthmani:
### Context

Now that we are moving the rate of pay and continuous payment control into a global setting, we need a way to allow users to disable WM if they choose to. 

Related to https://github.com/interledger/web-monetization-extension/issues/568

### Todos

* Design the best way to disable WM 
* Implement design 

### Acceptance Criteria 

- An Enable/Disable toggle will be available in the top right corner next to the Settings button.
- When WM is disabled, the following text will show on the screen : **'It looks like Web Monetization is disabled. To enable it, please adjust the toggle (icon) at the top right of the popup.'**

![WM enable-disable](https://github.com/user-attachments/assets/392dbf53-5329-4e8a-b6df-cb9144cb7ef8)

melissahenderson:
I have a suggestion for a shorter string: 

> It looks like you've disabled Web Monetization. To enable it, click the toggle in the upper right.

A few thoughts, for what they're worth...

- "It looks like Web Monetization is disabled." Is it possible that this message is misunderstood as meaning the site isn't web monetized? I guess anything and everything is possible ;)
- I know saying "please" is polite, but I don't think it's necessary here. Just my opinion.
- I think most people know what a toggle is. I feel strongly that "(icon)" isn't necessary.
- Semantics here, but top-right is where the settings cog is. Either way, if most people know what a toggle is, they won't be confused :)
- I don't think we should use the term "popup". To see the message, the user is already where they need to be. Other than top right or upper right, I don't think we need to clarify where they are in the UI.

PR #727, opened 2024-11-25
refactor(bg/storage): use separate key to toggle continuous payments

sidvishnoi:
## Context

Will help with https://github.com/interledger/web-monetization-extension/issues/623 as decided in https://github.com/interledger/web-monetization-extension/pull/654#issuecomment-2407142584

Then in future `enabled` will mean entire extension enabled or not, not just continuous payments.

- Rename `Storage.enabled` to `Storage.continuousPaymentsEnabled`
- Migrate extension storage from version 3 to 4
  - set `continuousPaymentsEnabled` to previous value of `enabled`
  - set `enabled` to `true` for [#623](https://github.com/interledger/web-monetization-extension/issues/623)
- Update entire codebase to reflect this rename

## Notes to reviewer

- A shorter name suggestion is definitely welcome.


src/background/services/storage.ts
```diff
@@ -284,4 +286,9 @@ const MIGRATIONS: Record<Storage['version'], Migration> = {
     data.state = newState satisfies Storage['state'];
     return [data];
   },
+  4: (data) => {
+    data.continuousPaymentsEnabled = data.enabled === true;
```
raducristianpopa, line 290:
```suggestion
    data.continuousPaymentsEnabled = data.enabled
```

raducristianpopa (commented):
Now thinking about it, we might not even need to delete the `enabled` key now, if we plan to introduce the global `enabled` flag. We can just keep it for when that lands.

sidvishnoi:
Yes, but its value isn't same as current meaning of enabled. We'll add it back (as new migration) with its new meaning.

sidvishnoi:
Update: oh yeah we can keep it. Just not use its value or set it yet.

DarianM:
LGTM

Issue #710, opened 2024-11-12
Replace rate of pay slider with InputAmount


sidvishnoi:
InputAmount needs inc/dec controls. See https://github.com/interledger/web-monetization-extension/issues/568#issuecomment-2399720161 for design and acceptance criteria reference.

sidvishnoi:
@arashveysi3 Or if you prefer a more frontend focused issue, this would be a good one.

PR #736, opened 2024-11-26
feat(Settings/Rate): replace slider with numeric input

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/710

- Add inc/dec buttons to InputAmount when `controls=true`
- Inc/dec with arrow keys and in larger step with `Shift + arrow keys` or PageUp/PageDown
- Replace Slider in Settings/Rate with InputAmount
- Refactor InputAmount to reduce re-renders, use memo and callback hook in more places


sidvishnoi:
@RabebOthmani

src/popup/components/ui/Input.tsx
```diff
@@ -58,11 +61,17 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
...
+      'Either addOn or addOnRight should be provided, but not both.',
+    );
+  }
+
   return (
     <div className="space-y-2">
       {label ? <Label htmlFor={id}>{label}</Label> : null}
       {description ? <p className="px-2 text-xs">{description}</p> : null}
-      <div className="relative">
+      <div className="relative w-fit">
```
DarianM, line 74: `width: fit-content` breaks other inputs
image

src/popup/components/Settings/RateOfPay.tsx
```diff
@@ -121,31 +117,38 @@ const RateOfPayInput = ({
...
+        onChange={(rate) => {
+          setErrorMessage('');
+          onRateChange(rate);
+        }}
+        onError={(error) => setErrorMessage(t(error))}
+        errorMessage={errorMessage}
+        min={Number(formatAmount(minRateOfPay))}
+        max={Number(formatAmount(maxRateOfPay))}
+        amount={rateOfPay}
+        controls={true}
```
DarianM, line 149: no `controls` for `Budget` input amount as well?

raducristianpopa: I would say only for the rate of pay. The controls do not really fit normal input amounts (`amount` field in the setup screen or in the `budget` tab)

src/popup/components/InputAmount.tsx
```diff
@@ -44,67 +44,194 @@ export const InputAmount = ({
...
-      const value = input.value;
+  const inputRef = React.useRef<HTMLInputElement>(null);
+
+  const formatAmount = React.useCallback(
+    (value: number) => formatNumber(value, assetScale),
+    [assetScale],
+  );
+
+  const handleValue = React.useCallback(
+    (value: string, skipSetValue = false) => {
```
raducristianpopa, line 61: What are the exact cases when we want to skip the value?

src/popup/components/InputAmount.tsx
```diff
@@ -44,67 +44,194 @@ export const InputAmount = ({
...
+  const controlDec = React.useCallback(() => {
+    incOrDec(inputRef.current!, -1, step, formatAmount, handleValue, min, max);
+  }, [step, formatAmount, handleValue, min, max]);
+
   return (
     <Input
       id={id}
       type="text"
       inputMode="numeric"
+      ref={inputRef}
```
raducristianpopa, line 130: Since `type=number` implicitly has the role of `spinbutton`, but we are using `type=text`.
```suggestion
      role="spinbutton"
      ref={inputRef}
```

src/popup/components/InputAmount.tsx
```diff
@@ -113,14 +240,42 @@ export function validateAmount(
...
+
+  React.useEffect(() => stop, [callback, ms]);
+
+  return {
+    onMouseDown: start,
+    onMouseUp: stop,
+    onMouseLeave: stop,
+  };
+}
+
```
raducristianpopa, line 278: Lets move the hooks into `lib/hooks.tsx` as well.

src/popup/components/ui/Input.tsx
```diff
@@ -36,20 +36,25 @@ export interface InputProps
   readOnly?: boolean;
   addOn?: React.ReactNode;
   addOnPosition?: 'left' | 'right';
+  addOnRight?: React.ReactNode;
```
raducristianpopa, line 39: Instead of having the `addOnPosition`, I would suggest to have diferent namings for the additional content:
* `leadingAddOn`: if it needs to be placed before the `input` component
* `trailingAddOn`: if it needs to be placed after the `input` component

raducristianpopa (commented):
There is a small edge case in `ConnectWallet`.

image

image


It looks like the amount field is still editable even if the wallet address field is incorrect or when the wallet address field is still not "dirty" - the user did not even focus the field.

src/popup/components/InputAmount.tsx
sidvishnoi: When user is typing, for example ”1", it'll get set formatted to ”1.00” with cursor moved to end on input field. So to avoid that, we skip that call. Inc/dec don't require such cursor, so ok (and even better) to format then.

src/popup/components/InputAmount.tsx
sidvishnoi: Should add this only when controls is true though. Will do.

sidvishnoi:
> There is a small edge case in `ConnectWallet`. https://github.com/interledger/web-monetization-extension/pull/736#pullrequestreview-2469166026

@raducristianpopa  Unrelated to this PR. But on paste, if URL looks valid, we shift focus automatically to amount field, even though it's readonly at the moment. This way, user will be able to edit amount right after paste (assuming the timings are right). The validation for URL is valid wallet address or not runs in background.

But I'll remove the readonly part from that input altogether (as separate PR). It's not right.

raducristianpopa (approved):

src/popup/components/Settings/RateOfPay.tsx
RabebOthmani: Agree with Radu! Usually the user has a clear idea about the amount , an input field makes sense. The rate of pay on the other hand is small. 
@sidvishnoi This looks good. No changes required for the amount.

RabebOthmani (approved):

Issue #743, opened 2024-11-29
Register keyAutoAdd content scripts dynamically


sidvishnoi:
Instead of updating manfiest.json every time there's a new wallet to support key auto add, keep entire logic related to that in `background/services/keyAutoAdd.ts`

PR #745, opened 2024-11-29
refactor: dynamically register key auto-add scripts

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/743

- Add a `KeyAutoAddService.registerContentScripts` to register key auto-add scripts
  - Don't register if already registered
  - Register in `background.start()`
- Remove key add content scripts from manifest.json
- Now all wallet-specific & key-add-specific URLs are in single place.


src/background/services/keyAutoAdd.ts
```diff
@@ -178,25 +178,49 @@ export class KeyAutoAddService {
       return false;
     }
   }
+
+  static async registerContentScripts({ browser }: Pick<Cradle, 'browser'>) {
+    const scripting = browser.scripting;
```
DarianM, line 183: nitpick: can use destructuring here

src/background/services/keyAutoAdd.ts
```diff
@@ -178,25 +178,49 @@ export class KeyAutoAddService {
       return false;
     }
   }
+
+  static async registerContentScripts({ browser }: Pick<Cradle, 'browser'>) {
```
raducristianpopa, line 182: I would make this a non-static method, since it is already registered in the container, to avoid passing around the `browser` instance. We will have to add the service in the background if we do this.

src/background/services/keyAutoAdd.ts
```diff
@@ -178,25 +178,49 @@ export class KeyAutoAddService {
...
+function getContentScripts(): Scripting.RegisteredContentScript[] {
+  return [
+    {
+      id: 'keyAutoAdd/testWallet',
+      matches: [
+        'https://wallet.interledger-test.dev/*',
+        'https://wallet.interledger.cards/*',
+      ],
+      js: ['content/keyAutoAdd/testWallet.js'],
+      runAt: 'document_end',
```
raducristianpopa, line 202: Any particular reason we are doing this at `document_end` ?

sidvishnoi: Oh, `document_idle` is a better choice. I though end was fired later 🤦🏽

DarianM:
Protect Document

DarianM (commented):
test-e2e

raducristianpopa (approved):

PR #747, opened 2024-12-02
refactor(tests/e2e): extract `getContinueWaitTime` helper

sidvishnoi:
Only difference was in defaultWaitMs, which we now pass as parameter.

### DIFF
--- a/tests/e2e/helpers/testWallet.ts
+++ b/tests/e2e/helpers/testWallet.ts
@@ -6,12 +6,12 @@ import {
   type KeyInfo,
 } from '../fixtures/helpers';
 import { fillPopup, type Popup, type ConnectDetails } from '../pages/popup';
-import { getWalletInformation } from '@/shared/helpers';
-import { waitForWelcomePage } from './common';
+import { getContinueWaitTime, waitForWelcomePage } from './common';
 
 export const KEYS_PAGE_URL = `https://wallet.interledger-test.dev/settings/developer-keys`;
 export const LOGIN_PAGE_URL = `https://wallet.interledger-test.dev/auth/login?callbackUrl=%2Fsettings%2Fdeveloper-keys`;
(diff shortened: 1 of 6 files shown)


sidvishnoi (commented):
test-e2e

raducristianpopa (approved):

tests/e2e/helpers/testWallet.ts
```diff
@@ -6,12 +6,12 @@ import {
...
 } from '../fixtures/helpers';
 import { fillPopup, type Popup, type ConnectDetails } from '../pages/popup';
-import { getWalletInformation } from '@/shared/helpers';
-import { waitForWelcomePage } from './common';
+import { getContinueWaitTime, waitForWelcomePage } from './common';
 
 export const KEYS_PAGE_URL = `https://wallet.interledger-test.dev/settings/developer-keys`;
 export const LOGIN_PAGE_URL = `https://wallet.interledger-test.dev/auth/login?callbackUrl=%2Fsettings%2Fdeveloper-keys`;
 export const API_URL_ORIGIN = `https://api.wallet.interledger-test.dev`;
+export const DEFAULT_CONTINUE_WAIT_MS = 1000;
```
raducristianpopa, line 14: If you can remind me: why do we have different default wait times for certain wallets?

sidvishnoi: That's what wallets chose really. I'm not sure where they configure it in Rafiki though.

raducristianpopa: > That's what wallets chose really. I'm not sure where they configure it in Rafiki though.

It's an environment variable - if it is not provided, it will default to 5 seconds.

Based on this, would it make sense to have the global default as 5 seconds as well? 🤔 

This way we won't have to configure a default one for each provider.

sidvishnoi: Makes sense.

raducristianpopa (approved):

PR #757, opened 2024-12-03
fix: don't try to localize appName in manifest.json

sidvishnoi:
## Context

While uploading to Chrome Webstore last time (with Preview suffix in name), it displayed name as `__MSG_appName__ Preview`. I wasn't sure if it'll parse name correctly, but I didn't take risk and updated name in manifest manually and uploaded the new zip.

Besides, `name` is something we won't be localizing, so it doesn't make sense to add it in locales.

The only loss I see is not having all strings within locales. I think might be fine for this specific case.

- Remove `appName` key from `_locales/en/messages.json`
- Add name directly to `manifest.json`
- Get name where we need it in extension from `browser.runtime.getManifest()`, instead of `t('appName')`.

> [!TIP]
> We get added benefit that now we'll show full name (with Nightly or Preview suffix) when hovering over extension icon or with key nickname, to help differentiate when user has multiple versions installed.

Tested working in Firefox and Chrome.


src/background/services/tabEvents.ts
```diff
@@ -179,7 +179,7 @@ export class TabEvents {
     state: Storage['state'];
     tabInfo: PopupTabInfo;
   }) {
-    let title = this.t('appName');
+    let { name: title } = this.browser.runtime.getManifest();
```
sidvishnoi, line 182: I wonder if I should the name (as `appName`) in `Cradle` to ensure we don't have to getManifest repeatedly?

src/background/services/keyAutoAdd.ts
```diff
@@ -38,10 +39,11 @@ export class KeyAutoAddService {
   constructor({
     browser,
     storage,
+    appName,
     browserName,
     t,
-  }: Pick<Cradle, 'browser' | 'storage' | 'browserName' | 't'>) {
-    Object.assign(this, { browser, storage, browserName, t });
+  }: Pick<Cradle, 'browser' | 'storage' | 'appName' | 'browserName' | 't'>) {
```
raducristianpopa, line 45: No need to use `Pick` since whole cradle is passed by default when using `PROXY_MODE`.
```suggestion
  }: Cradle) {
```

raducristianpopa (approved):

src/background/services/keyAutoAdd.ts
sidvishnoi: Would've to manually pass each param when instantiating `KeyAutoAdd` (not using awilix there, it's not on container).

raducristianpopa: 🤦

PR #760, opened 2024-12-05
refactor(popup): use valtio external store for state management

sidvishnoi:
Uses valtio for better render performance, and moves state out of context into a store in file.

- Turn `PopupContextProvider` to `WaitForStateLoad` to initialize store.
- Keep `usePopupState` as helper, even though it won't return dispatch now (only readonly state will be returned)
- Move state related things from `context.tsx` to `store.tsx`.
- Improve performance of RateOfPay screen as a test (we had too many props, not enough state)
- Kept `dispatch` mechanism for actions to update state. We don't need to change this right away.
- Tried to keep PR as small as possible. Most files just ±N line changes.

Adds +3KB to zip and +6KB to popup.js

### DIFF
--- a/src/pages/popup/components/ConnectWalletForm.tsx
+++ b/src/pages/popup/components/ConnectWalletForm.tsx
@@ -20,7 +20,7 @@ import {
 } from '@/shared/helpers';
 import type { WalletAddress } from '@interledger/open-payments';
 import type { ConnectWalletPayload, Response } from '@/shared/messages';
-import type { PopupTransientState } from '@/shared/types';
+import type { DeepReadonly, PopupTransientState } from '@/shared/types';
 
 interface Inputs {
   walletAddressUrl: string;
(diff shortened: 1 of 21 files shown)


src/pages/popup/components/ConnectWalletForm.tsx
```diff
@@ -29,14 +29,15 @@ interface Inputs {
   autoKeyAddConsent: boolean;
 }
 
+type ConnectTransientState = DeepReadonly<PopupTransientState['connect']>;
```
DarianM, line 32: just to validate my thoughts, `DeepReadonly` to prevent unintended mutations to connect state inside `ConnectWalletForm` component?

DarianM (approved):
pretty neat 💯

raducristianpopa (approved):
LGTM! 🚀

Issue #558, opened 2024-08-30
Improve post-install flow


sidvishnoi:
Someone (Santosh?) mentioned in yesterday's community call, that extension by default gets hidden in extension list and they need to pin it. Someone also suggested having a post-install page that tells user what's next.

I installed 1password extension today, and I think we can take a lot of inspiration here:
Their post-install page is an extension page itself, `chrome-extension://aeblfdkhhhdcdjpifhhbdiojplfjncoa/app/app.html#/page/ready` so it knows more easily about the extension state. (We had to use http://webmonetization.org/welcome on post-connect screen as OpenPayments redirect had to be `https://`)

![image](https://github.com/user-attachments/assets/41548c3d-775b-4a98-839b-343b2042fc37)

TODO:
- [ ] Add post-install page with next-steps
- [ ] Recommend pinning extension for easier access
- [ ] Anything else we want there?

RabebOthmani:
Melissa is working on a getting started page in the documentation so we don't need to invent the wheel. We will need to add learn more links to the extension and make sure that the doc's page is our post install page. 
I tried a bunch of extensions to see the experience. Some create new pages within the extension, others link to pages from within their websites

RabebOthmani:
@sidvishnoi @raducristianpopa based on last week's call and before working on the content for a post-install HTML page, I'd like to confirm with you: do we think creating an HTML page within the extension is better than linking to a getting started page on the WM website?

sidvishnoi:
Depends what we want to show on it:
- Documentation? Best on separate page.
- Using the extension (like screenshot in this issue's description): within extension

RabebOthmani:
I understand that there are both flavours of a post install 'welcome page'. The objective is to try to educate users early on on things like how to get a wallet, understand how to overall get started with the extension, etc. which are things that are already in a getting started page ready to go on WM website. My assumption and based on a face to face conversation with @raducristianpopa during the Rafiki work week, is that we will utilize this getting started page. 
Last week during our team call, I got the impression that it might be an issue going that route and that's what I wanted to clarify.

PR #761, opened 2024-12-05
feat(pages/app): add boilerplate for page with post-install screen

sidvishnoi:
## Context

Part of https://github.com/interledger/web-monetization-extension/issues/558

- Add new `app` extension page boilerplate, with router, messaging and state management set up.
- No layout yet.

### DIFF
--- a/src/pages/app/App.tsx
+++ b/src/pages/app/App.tsx
@@ -1,0 +1,5 @@
+import React from 'react';
+
+export function App() {
+  return <h1>Hello</h1>;
+}
--- a/src/pages/app/App.tsx
+++ b/src/pages/app/App.tsx
@@ -1,5 +1,45 @@
 import React from 'react';
+import {
+  BrowserContextProvider,
+  TranslationContextProvider,
+} from '@/pages/shared/lib/context';
+import browser from 'webextension-polyfill';
+import {
+  createMemoryRouter,
+  RouterProvider,
+  type RouteObject,
+} from 'react-router-dom';
+import { MessageContextProvider } from './lib/context';
 
-export function App() {
-  return <h1>Hello</h1>;
-}
+export const ROUTES_PATH = {
+  HOME: '/',
+  POST_INSTALL: '/post-install',
+} as const;
+
+export const routes = [
+  {
+    children: [
+      {
+        path: ROUTES_PATH.HOME,
+        lazy: () => import('./pages/Home'),
+      },
+      {
+        path: ROUTES_PATH.POST_INSTALL,
+        lazy: () => import('./pages/PostInstall'),
+      },
+    ],
+  },
+] satisfies RouteObject[];
+
+export const App = () => {
+  const router = createMemoryRouter(routes);
+  return (
+    <BrowserContextProvider browser={browser}>
+      <TranslationContextProvider>
+        <MessageContextProvider>
+          <RouterProvider router={router} />
+        </MessageContextProvider>
+      </TranslationContextProvider>
+    </BrowserContextProvider>
+  );
+};
--- a/src/pages/app/App.tsx
+++ b/src/pages/app/App.tsx
@@ -11,7 +11,7 @@ import {
 } from 'react-router-dom';
 import { MessageContextProvider } from './lib/context';
 
-export const ROUTES_PATH = {
+export const ROUTES = {
   HOME: '/',
   POST_INSTALL: '/post-install',
 } as const;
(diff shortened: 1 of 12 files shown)


src/pages/app/index.css
```diff
@@ -0,0 +1,37 @@
+@tailwind base;
```
sidvishnoi, line 1: We can reuse this CSS across files, but it'd require a postcss setup. Can add in future, for now I was keeping it light and let TailwindCSS manage it from defaults.

src/pages/app/pages/Home.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const Component = () => {
+  const state = useAppState();
+  return (
+    <div className="mx-auto max-w-2xl space-y-4 p-6">
+      <div className="flex items-center justify-between">
+        <h1 className="text-2xl">Home</h1>
+        <nav>
+          <ul>
+            <li>
+              <Link
```
DarianM, line 15: this puzzles me a bit, can't we show the postinstall page directly?
why i need to click an additional link?
or is this temporary ?

sidvishnoi: This is temporary. Will remove, and keep it as simple as possible in this PR.

sidvishnoi: We'll have multiple screens in this `/app/index.html`, post-install being one of them. Rest unknown at the moment, but we'll have more!

src/pages/app/App.tsx
```diff
@@ -0,0 +1,41 @@
...
+import browser from 'webextension-polyfill';
+import {
+  createMemoryRouter,
+  RouterProvider,
+  type RouteObject,
+} from 'react-router-dom';
+import { MessageContextProvider, WaitForStateLoad } from './lib/context';
+
+export const ROUTES = {
+  HOME: '/',
```
raducristianpopa, line 15: What is the home route being used for?

sidvishnoi: Unused for now, just there for router default page. I think Home will correspond to a Settings layout in future (/home/setting1, /home/setting2).
Logic for showing what route we want to render will come in follow-up PRs as well implement them. For example, if connected once already, showing post install doesn't make sense again - so Home would probably just show a message like "use popup to control extension" until Home can have some actual content.

sidvishnoi:
> I feel like naming this `app` is a bit cumbersome. I would probably go for something like `extras`, where we can have multiple screens if needed.

@raducristianpopa Used `app` inspired by 1Password:
image

`extras` feels a bit not-right. Note that, what name we use will be visible in URL bar when extension page is opened (right now from post-install only; but in future more sources to open it). Can have different name in src and output btw (but that might add to more confusion?). `app` feels it covers everything.

Also note that, we don't want multiple `.html` pages - as each adds like 100KB to extension zip. Using that one app/extras for everything is better here.

DarianM (approved):

src/background/services/keyAutoAdd.ts
```diff
@@ -78,11 +82,17 @@ export class KeyAutoAddService {
     return this.tab?.id;
   }
 
-  private async process(url: string, payload: BeginPayload) {
+  private async process(url: string, payload: BeginPayload, tabId?: TabId) {
     const { resolve, reject, promise } = withResolvers();
 
-    const tab = await this.browser.tabs.create({ url });
+    await this.browser.tabs.remove(tabId!);
```
sidvishnoi, line 88: TODO: remove this stray remove call.

src/assets/images/wallet-signup-1-chrome.png
raducristianpopa, line 1: Why do we need 3 images for the sign up form? There is no need, we can only use one image for all browsers.

raducristianpopa (commented):
Can we open the next step when the current one is closed?

src/assets/images/wallet-signup-1-chrome.png
sidvishnoi: I think we did that as we were showing URL to interledger.app, so matched browser UI there.

src/pages/app/pages/PostInstall.tsx
```diff
@@ -0,0 +1,220 @@
...
+  return (
+    <ol className="flex flex-col gap-4">
+      <Step
+        index={0}
+        open={isOpen === 0}
+        onClick={onClick}
+        title={
+          <>
+            {t('postInstall_text_stepGetWallet_title')}{' '}
+            <a
```
DarianM, line 88: href destination does not work on click

src/pages/app/pages/PostInstall.tsx
```diff
@@ -0,0 +1,220 @@
...
+  }, []);
+
+  return (
+    <ol className="flex flex-col gap-4">
+      <Step
+        index={0}
+        open={isOpen === 0}
+        onClick={onClick}
+        title={
+          <>
```
DarianM, line 86: really smol nitpick: React.Fragment instead of `<>` or vice-versa for consistency

DarianM (approved):
LGTM 🚀

raducristianpopa (approved):

Issue #471, opened 2024-08-02
Cannot enable continuous payments on non-monetized screen


sidvishnoi:
Need to go to monetized websites to configure the settings. Quite annoying.

<img src="https://github.com/user-attachments/assets/2b4aedda-743f-44d8-8791-cf9a16ce21c5" width="200"/>

PR #773, opened 2024-12-12
feat: add global payment switch

sidvishnoi:
## Context

- Closes https://github.com/interledger/web-monetization-extension/issues/623
- Closes https://github.com/interledger/web-monetization-extension/issues/471
- Closes https://github.com/interledger/web-monetization-extension/pull/654 (supersedes)


sidvishnoi:
The toggle doesn't look right here, and it's not clear what the toggle would do. Will add power icon.

src/background/services/storage.ts
```diff
@@ -116,14 +116,6 @@ export class StorageService {
...
   }
 
-  async getWMState(): Promise<boolean> {
-    const { continuousPaymentsEnabled } = await this.get([
-      'continuousPaymentsEnabled',
-    ]);
-
-    return continuousPaymentsEnabled;
-  }
-
```
sidvishnoi, line 126: This was unused.

src/background/services/monetization.ts
```diff
@@ -257,12 +275,27 @@ export class MonetizationService {
   }
 
   async toggleWM() {
```
raducristianpopa, line 277:
```suggestion
  async toggleContinuousPayments() {
```

src/pages/popup/components/TogglePaymentsButton.tsx
raducristianpopa, line 1: Any particular reason for having two components for this? I think we can just merge them into `TogglePaymentsButton`.

raducristianpopa (commented):
As a safety measure, we can check if payments are enabled in the `MonetizationService.pay` as well.

sidvishnoi (commented):
test-e2e

raducristianpopa (approved):

src/background/services/monetization.ts
```diff
@@ -256,13 +274,28 @@ export class MonetizationService {
...
-    const { continuousPaymentsEnabled } = await this.storage.get([
+  async toggleContinuousPayments() {
+    const { continuousPaymentsEnabled, enabled } = await this.storage.get([
       'continuousPaymentsEnabled',
+      'enabled',
     ]);
     const nowEnabled = !continuousPaymentsEnabled;
     await this.storage.set({ continuousPaymentsEnabled: nowEnabled });
-    if (nowEnabled) {
+    if (nowEnabled && enabled) {
```
DarianM, line 284: qq: why do we have to check both `continuousPaymentsEnabled` and `enabled` here? 🤔 
`enabled` is the bool for the entire extension monetization, right?

sidvishnoi: If user has disabled continuous payments from toggle in Settings > Rate, but enabled is true, payments are still not to get through.

DarianM (approved):

PR #779, opened 2024-12-16
fix(background/KeyAutoAdd): invalid tabId in Firefox

### LINKED ISSUE #778: [BUG] uncaught invalid tabId for auto key addition; Firefox
### Steps to reproduce

connect wallet using Firefox (ensure public key does not already exist)
Image

### Expected result

should properly connect the wallet whenever the public key does not already exist

### Actual result

auto key add service throws error whenever:
- not triggered from post-install screen
- key not already added

### Screenshots or videos

_No response_

### Additional context

`addPublicKeyToWallet` process throws error when `tabId` is undefined.
Image
Firefox does not support `-1` for `await this.browser.tabs.get(tabId ?? -1)`, meanwhile chrome does not throw using `-1` as id.

for more context, `await this.browser.tabs.get(tabId ?? -1)` throws synchronously in Firefox:
Image
meanwhile chrome:
Image

### Operating system

Windows

### Operating system version

_No response_

### Browsers

Firefox

### Browser version

_No response_

### Extension version

0bf9d6d3f57ebc11ba3d992bcee272a604ebf23d

DarianM:
## Context
`this.browser.tabs.get()` can throw a synchronous error before returning a promise.

If `this.browser.tabs.get(tabId ?? -1)` throws a synchronous error before returning a promise (Promise.reject), the .catch() won’t catch it, as the error occurs before the promise is created.

Closes #778 

## Proposed change
- use a `try-catch` block to handle  `browser.tabs.get` exception

### DIFF
--- a/src/background/services/keyAutoAdd.ts
+++ b/src/background/services/keyAutoAdd.ts
@@ -91,9 +91,13 @@ export class KeyAutoAddService {
     existingTabId?: TabId,
   ) {
     const { resolve, reject, promise } = withResolvers();
-    const tab = existingTabId
-      ? await this.browser.tabs.update(existingTabId, { url })
-      : await this.browser.tabs.create({ url });
+    let tab: Tabs.Tab;
+    try {
+      tab = await this.browser.tabs.get(existingTabId ?? -1);
+      await this.browser.tabs.update(tab.id, { url });
+    } catch {
+      tab = await this.browser.tabs.create({ url });
+    }
 
     this.tab = tab;
     if (!tab.id) {
--- a/src/background/services/keyAutoAdd.ts
+++ b/src/background/services/keyAutoAdd.ts
@@ -6,14 +6,15 @@ import {
   withResolvers,
   type ErrorWithKeyLike,
 } from '@/shared/helpers';
-import type { Browser, Runtime, Scripting, Tabs } from 'webextension-polyfill';
+import type { Browser, Runtime, Scripting } from 'webextension-polyfill';
 import type { WalletAddress } from '@interledger/open-payments';
-import type { TabId } from '@/shared/types';
+import type { Tab, TabId } from '@/shared/types';
 import type { Cradle } from '@/background/container';
 import type {
   BeginPayload,
   KeyAutoAddToBackgroundMessage,
 } from '@/content/keyAutoAdd/lib/types';
+import { reuseOrCreateTab } from '../utils';
 
 export const CONNECTION_NAME = 'key-auto-add';
 
@@ -34,7 +35,7 @@ export class KeyAutoAddService {
   private browserName: Cradle['browserName'];
   private t: Cradle['t'];
 
-  private tab: Tabs.Tab | null = null;
+  private tab: Tab | null = null;
 
   constructor({
     browser,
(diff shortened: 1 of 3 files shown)


src/background/services/keyAutoAdd.ts
```diff
@@ -82,13 +85,16 @@ export class KeyAutoAddService {
...
 
-  private async process(url: string, payload: BeginPayload, tabId?: TabId) {
+  private async process(
+    url: string,
+    payload: BeginPayload,
+    existingTabId?: TabId,
+  ) {
     const { resolve, reject, promise } = withResolvers();
+    const tab = existingTabId
+      ? await this.browser.tabs.update(existingTabId, { url })
```
sidvishnoi, line 95: Only concern with this approach is, what happens if tab with this ID is closed? TabId will be defined, but update call will throw.

DarianM: my bad, now I understand why `tabs.get` was used there.
let's use a `try catch` to handle that sync error that tabs.get method can throw

src/background/services/keyAutoAdd.ts
```diff
@@ -82,13 +85,20 @@ export class KeyAutoAddService {
...
+    existingTabId?: TabId,
+  ) {
     const { resolve, reject, promise } = withResolvers();
+    let tab: Tabs.Tab;
+    try {
+      tab = await this.browser.tabs.get(existingTabId ?? -1);
+      await this.browser.tabs.update(tab.id, { url });
+    } catch {
+      tab = await this.browser.tabs.create({ url });
+    }
```
sidvishnoi, line 100: Let's extract this into a helper function, `reuseOrCreateTab(browser, url, tabId?): Promise<Tab>` in `background/utils`.

DarianM: if extract, let's also use `reuseOrCreateTab` in `getInteractionInfo` now. what do you think?

src/background/services/keyAutoAdd.ts
```diff
@@ -82,81 +86,82 @@ export class KeyAutoAddService {
...
 
-  private async process(url: string, payload: BeginPayload, tabId?: TabId) {
+  private async process(
+    url: string,
+    payload: BeginPayload,
+    existingTabId?: TabId,
+  ): Promise<unknown> {
     const { resolve, reject, promise } = withResolvers();
+    try {
+      const tab = await reuseOrCreateTab(this.browser, url, existingTabId);
```
sidvishnoi, line 96: Let it throw. It's unexpected and should not happen. No need a try-catch for `reuseOrCreateTab`.

sidvishnoi (changes requested):
I promise it's last change request.

raducristianpopa (approved):

src/background/utils.ts
```diff
@@ -86,6 +86,26 @@ export const getTab = (sender: Runtime.MessageSender): Tab => {
...
+  browser: Browser,
+  url: string,
+  tabId?: number,
+): Promise<Tab> => {
+  try {
+    const tab = await browser.tabs.get(tabId ?? -1);
+    if (!tab.id) {
+      throw new Error('Could not retrieve tab.');
+    }
+    return (await browser.tabs.update(tab.id, { url })) as Tab;
```
sidvishnoi, line 99:
```suggestion
    let tab = await browser.tabs.get(tabId ?? -1);
    if (!tab.id) {
      throw new Error('Could not retrieve tab.');
    }
    tab = await browser.tabs.update(tab.id, { url });
    return tab as Tab;
```

PR #785, opened 2024-12-17
refactor(background): remove OP error map as unused

sidvishnoi:


DarianM (approved):
I know we catch invalid client throughout different bindMessage cases now and we use `error.code === 'invalid_client'` instead of e.description for invalid client,  so this cleanup LGTM

PR #792, opened 2024-12-18
fix(Popup): update global payment toggle UI

### LINKED ISSUE #787: [UX feedback]: global payment toggle
Based on feedback during yesterday's team meeting, update global payment toggle feature's UI.

- Remove the large power button on Home screen when disabled. Use `NotMonetized` component to display message.
- Continue showing the button in Header, don't remove it to show the large button.
- Show the default colored icon in toolbar, instead of disabled gray one.

DarianM:
## Context

Closes #787 .

- Removed the large power button on Home screen when disabled, display disable message only.
- Always showing the toggle button in Header.
- Show the default colored icon in toolbar, instead of disabled gray one.

### DIFF
--- a/src/pages/popup/components/TogglePaymentsButton.tsx
+++ b/src/pages/popup/components/TogglePaymentsButton.tsx
@@ -2,8 +2,7 @@ import React from 'react';
 import { Power } from '@/pages/shared/components/Icons';
 import { cn } from '@/shared/helpers';
 import { useMessage } from '@/popup/lib/context';
-import { dispatch } from '@/popup/lib/store';
-import { usePopupState } from '@/popup/lib/store';
+import { dispatch, usePopupState } from '@/popup/lib/store';
 
 export const TogglePaymentsButton = () => {
   const { enabled } = usePopupState();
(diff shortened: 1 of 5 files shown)


sidvishnoi (approved):
LGTM.
Maybe the toggle's `input` should've `onClick={(ev) => ev.currentTarget.blur()}` so the focus styles get removed?

src/pages/popup/components/TogglePaymentsButton.tsx
```diff
@@ -3,14 +3,10 @@ import { Power } from '@/pages/shared/components/Icons';
 import { cn } from '@/shared/helpers';
 import { useMessage } from '@/popup/lib/context';
 import { dispatch } from '@/popup/lib/store';
+import { usePopupState } from '@/popup/lib/store';
```
sidvishnoi, line 6:
```suggestion
import { useMessage, usePopupState } from '@/popup/lib/context';
import { dispatch } from '@/popup/lib/store';
```

sidvishnoi, line 6:
```suggestion
import { dispatch, usePopupState } from '@/popup/lib/store';
```

DarianM: should we add eslint to enforce a single import statement for modules with the same path? [no-duplicate-imports](https://eslint.org/docs/latest/rules/no-duplicate-imports)

Issue #655, opened 2024-10-10
Close popup in Firefox when connecting


sidvishnoi:
Popup closes in Chrome when new tab opens, but not in Firefox. Should close to bring user's attention to connect process (but only when the new tab opens).

devcer:
https://github.com/user-attachments/assets/2483ea89-9b7b-4eee-a47c-de7e220914df

Just tested the main branch and didn't seem to reproduce. any steps to reproduce this?

sidvishnoi:
@devcer it stays open in Firefox once you begin the connect wallet process.

sidvishnoi:
@arashveysi3 Do you want to try this one?

arashveysi3:
Looks great to me. I'll take this.

sidvishnoi:
@arashveysi3 Still interested in this issue?

PR #807, opened 2025-01-06
fix: close popup when tab navigates during connect

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/655

- Send a message `CLOSE_POPUP` to close the popup, triggered by `connect_wallet.close_popup` event.
   - Closes popup when tab opens/navigates due to key addition or consent approval.
- Remove `queue` from `SendToPopup`, as not needed and causes issues with above message.
   - `queue` was added when we weren't syncing tab state live into popup (we sync now with `SET_TAB_DATA`).
- The issue wasn't only with Firefox. When current reused tab has a navigation (like from `app/index.html` to wallet key-add page or to consent page), the popup didn't auto-close in Chrome either.
- E2E tests break due to us extensively relying on popup being open, so make `window.close()` a no-op in popup.


sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

src/background/services/sendToPopup.ts
```diff
@@ -25,10 +24,6 @@ export class SendToPopup {
       }
       this.port = port;
       this.isConnected = true;
-      for (const [type, data] of this.queue) {
```
DarianM, line 28: what was the initial use case/intent with this queue?
trying to understand why the cleanup now

DarianM (approved):

PR #839, opened 2025-01-16
perf: use global browser instead of webextension-polyfill

sidvishnoi:
## Context

We don't need the polyfill except for async message handlers in background.

Remove polyfilled browser from content scripts and React apps to reduce bundle size, and have less JS in content scripts for better performance.

Keeping it draft until we add more E2E tests.

### DIFF
--- a/src/content/container.ts
+++ b/src/content/container.ts
@@ -4,7 +4,7 @@ import {
   createContainer,
   InjectionMode,
 } from 'awilix/browser';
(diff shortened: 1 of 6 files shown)


sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

PR #872, opened 2025-01-28
refactor(PostInstall): use `ConnectWalletForm` in the last step

### LINKED ISSUE #793: Post-install: properly show connect wallet form in step
https://github.com/interledger/web-monetization-extension/pull/789 embeds popup as an iframe to show the connect wallet form. Integrate `ConnectWalletForm` component so it works natively & cleanly in post-install screen.

DarianM:
## Context
Closes #793 

- replace popup iframe with ConnectWalletForm in PostInstall app
- add valtio store to app
- persist the shared state and sync it between both app and popup contexts
- integrate message passing between app <-> background <-> popup


src/background/services/sendToApp.ts
sidvishnoi, line 1: Instead of duplicating this, better create two instances with different `CONNECTION_NAME`, and make the class generic on message types.

sidvishnoi: Can name the class/file something like `MessagePort`

src/pages/app/lib/context.tsx
```diff
@@ -1 +1,70 @@
...
+
 export { useBrowser, useTranslation } from '@/pages/shared/lib/context';
+
+export function WaitForStateLoad({ children }: React.PropsWithChildren) {
+  const message = useMessage();
+  const [isLoading, setIsLoading] = React.useState(true);
+
+  React.useEffect(() => {
+    async function get() {
+      const response = await message.send('GET_CONTEXT_DATA');
```
sidvishnoi, line 19: I think this should be a separate message, so app gets its own data, which could be different than popup's data.

src/background/container.ts
```diff
@@ -101,7 +108,20 @@ export const configureContainer = () => {
...
     message: asClass(MessageManager<BackgroundToContentMessage>).singleton(),
     tabEvents: asClass(TabEvents).singleton(),
-    sendToPopup: asClass(SendToPopup).singleton(),
+    sendToPopup: asClass(
+      class extends SendToPort<BackgroundToPopupMessage> {
+        constructor(cradle: Cradle) {
+          super(cradle, BACKGROUND_TO_POPUP_CONNECTION_NAME);
+        }
+      },
+    ).singleton(),
```
sidvishnoi, line 117: I think we should be able to use [`inject`](https://github.com/jeffijoe/awilix#per-module-local-injections) here instead of creating new class?

src/background/services/background.ts
```diff
@@ -286,6 +290,15 @@ export class Background {
 
             // endregion
 
+            // region App
+            case 'GET_APP_CONTEXT_DATA':
+              return success(
+                await this.monetizationService.getPopupData(
+                  await this.windowState.getCurrentTab(),
+                ),
```
sidvishnoi, line 298: We don't need all this data. Can add a new function (ok within background.ts ok) to `getAppData()`

src/background/services/background.ts
```diff
@@ -189,7 +193,7 @@ export class Background {
         try {
           switch (message.action) {
             // region Popup
-            case 'GET_CONTEXT_DATA':
+            case 'GET_POPUP_CONTEXT_DATA':
```
sidvishnoi, line 196: Maybe this?
```suggestion
            case 'GET_DATA_POPUP':
```

src/background/services/sendToPort.ts
```diff
@@ -1,24 +1,21 @@
 import type { Runtime } from 'webextension-polyfill';
-import {
-  BACKGROUND_TO_POPUP_CONNECTION_NAME as CONNECTION_NAME,
-  type BackgroundToPopupMessage,
-  type BackgroundToPopupMessagesMap,
-} from '@/shared/messages';
 import type { Cradle } from '@/background/container';
 
-export class SendToPopup {
+export class SendToPort<Message> {
```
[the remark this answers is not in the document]
sidvishnoi: Or something about like `MessagesMap`?

src/pages/app/pages/PostInstall.tsx
```diff
@@ -151,11 +161,24 @@ const Steps = () => {
...
+                localStorage?.getItem('connect.recurring') === 'true' || false,
+              amount: localStorage?.getItem('connect.amount') || undefined,
+              walletAddressUrl:
+                localStorage?.getItem('connect.walletAddressUrl') || undefined,
+              autoKeyAddConsent:
+                localStorage?.getItem('connect.autoKeyAddConsent') === 'true',
+            }}
+            saveValue={(key, val) => {
+              localStorage?.setItem(`connect.${key}`, val.toString());
+            }}
```
sidvishnoi, line 178: Can you check, if the localStorage of app & popup is shared? If not, if user connects from app, they won't have pre-filled data in popup next time, and that would be a problem.

sidvishnoi (changes requested):
Nearly there!

src/pages/app/pages/PostInstall.tsx
DarianM: app and popup sharing localStorage: entering any "amount" and toggling "recurring" during login from the app, keeps the amount and toggle input later in the budget tab of the popup

sidvishnoi: Perfect!

src/background/container.ts
DarianM: i like this a lot. definitely much more cleaner 👏

src/background/services/sendToPort.ts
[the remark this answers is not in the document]
DarianM: had some hurdles in setting up a proper type to be reflected in SendToPort `send` 💭 opened for a change here

src/background/container.ts
```diff
@@ -101,7 +106,12 @@ export const configureContainer = () => {
       })),
     message: asClass(MessageManager<BackgroundToContentMessage>).singleton(),
     tabEvents: asClass(TabEvents).singleton(),
-    sendToPopup: asClass(SendToPopup).singleton(),
+    sendToPopup: asClass(SendToPort, {
+      injector: () => ({ connectionName: BACKGROUND_TO_POPUP_CONNECTION_NAME }),
+    }).singleton(),
+    sendToApp: asClass(SendToPort, {
+      injector: () => ({ connectionName: BACKGROUND_TO_APP_CONNECTION_NAME }),
+    }),
```
sidvishnoi, line 114: Nit: Let's use `inject` instead of `injector` like we do on rest of registrations, for consistency.

```suggestion
    sendToPopup: asClass(SendToPort)
      .singleton()
      .inject(() => ({
        connectionName: BACKGROUND_TO_POPUP_CONNECTION_NAME,
      })),
    sendToApp: asClass(SendToPort)
      .singleton()
      .inject(() => ({
        connectionName: BACKGROUND_TO_APP_CONNECTION_NAME,
      })),
```

src/background/services/background.ts
```diff
@@ -373,4 +383,13 @@ export class Background {
       this.logger.error(error);
     }
   };
+
+  getAppData = async () => {
```
sidvishnoi, line 387:
```suggestion
  async getAppData() {
```

src/background/services/sendToPort.ts
```diff
@@ -1,24 +1,22 @@
...
   private browser: Cradle['browser'];
-
-  private isConnected = false;
   private port: Runtime.Port;
+  private isConnected = false;
 
-  constructor({ browser }: Cradle) {
+  constructor(
+    { browser }: Cradle,
+    private readonly connectionName: string,
```
sidvishnoi, line 12: I think we'd need to extend Cradle here.
```suggestion
    { browser, connectionName }: Cradle & { connectionName: string },
```

DarianM: agree! if we gonna use `inject`

src/pages/app/pages/PostInstall.tsx
```diff
@@ -150,13 +160,39 @@ const Steps = () => {
...
         title={t('postInstall_action_submit')}
       >
-        <div className="mx-auto h-popup w-popup overflow-hidden">
-          <iframe
-            loading="lazy"
-            src={popupUrl}
-            title="Connect your wallet"
-            className="h-popup w-popup border-none"
-          />
+        <div className="text-center mx-auto h-auto w-popup pt-4 pb-8 overflow-hidden">
```
sidvishnoi, line 163: Consider extracting this step into a separate component, say, `StepConnectWallet`.

sidvishnoi (commented):
test-e2e

sidvishnoi (approved):
LGTM. Great work!

Issue #873, opened 2025-01-29
Fix E2E test flakiness


sidvishnoi:
As discovered in https://github.com/interledger/web-monetization-extension/pull/864#issuecomment-2618578727, https://github.com/interledger/web-monetization-extension/pull/864#issuecomment-2618814231 and https://github.com/interledger/web-monetization-extension/pull/864#issuecomment-2619199740.

sidvishnoi:
After numerous failed attempts today trying to clear extension state after each file, it's best to isolate tests for now, and create new browser for each test (not per file) until Playwright supports not reusing worker or cleanup per file.
This would make tests much slower, but at least they won't be flaky.

Another alternative would be to let go of fixtures and setup & teardown at top level manually, but that would be more trouble. Once Playwright has file level scope, we can simply replace scope in fixtures then.

sidvishnoi:
For performance with isolation, we can then increase parallelism with more workers and enabling fullyParallel.

PR #864, opened 2025-01-27
test(e2e): add tests for budget edit

sidvishnoi:
## Context

Part of https://github.com/interledger/web-monetization-extension/issues/610

PR #879, opened 2025-01-30
test(e2e): avoid sharing browser instance across workers

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/873

Playwright reuses worker wherever possible, which means we can't cleanup browser state in `beforeAll`/`afterAll` hooks as they run at the end of worker. Playwright creates a new worker when test fails, so next test/retry uses fresh state. In our case, we had already connected wallet, and if next test tried to connect wallet again, they will timeout looking for the form. We could've verified & disconnected wallet as a step in each test, but that's not good either - as browser storage & pending API calls can still affect later tests.

I tried various workarounds to run cleanup process in `beforeAll` per file in https://github.com/interledger/web-monetization-extension/pull/864, but nothing gave a clean enough state.

Playwright doesn't have a `file` scope (https://github.com/microsoft/playwright/issues/34519); it has only `worker` (state gets reused across files) or per `test` (new setup in each test within file), so here we decided to use `worker` scope to sacrifice performance/efficiency to avoid test flakiness.

- Create new browser context for each test, isolating them
- Use `test.step` where it would be too expensive to create new browser instance of each sub-test (See `not-monetized.spec.ts`, ignore whitespace changes)
- For better performance, increased parallelism for tests by not restricting to 1 worker, and enabling `fullyParallel`.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -9,10 +9,9 @@ if (!process.env.CI) {
 export default defineConfig({
   testDir,
   outputDir: path.join(testDir, 'test-results'),
-  fullyParallel: false,
+  fullyParallel: true,
   forbidOnly: !!process.env.CI,
   retries: process.env.CI ? 2 : 0,
-  workers: process.env.CI ? 1 : undefined,
   reporter: [
     ['list'],
     [
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -9,11 +9,6 @@ if (!process.env.CI) {
 export default defineConfig({
   testDir,
   outputDir: path.join(testDir, 'test-results'),
-  // We don't want this set to true as that would make tests in each file to run
-  // in parallel, which will cause conflicts with the "global state". With this
-  // set to false and workers > 1, multiple test files can run in parallel, but
-  // tests within a file are run at one at a time. We make extensive use of
-  // worker-scope fixtures and beforeAll hooks to achieve best performance.
   fullyParallel: false,
   forbidOnly: !!process.env.CI,
   retries: process.env.CI ? 2 : 0,
--- a/tests/e2e/connect.spec.ts
+++ b/tests/e2e/connect.spec.ts
@@ -1,10 +1,5 @@
-/// <reference types="chrome"/>
 import { test, expect } from './fixtures/base';
 import { connectWallet, disconnectWallet } from './pages/popup';
-
-test.beforeEach(async ({ popup }) => {
-  await popup.reload();
-});
 
 test('connects with correct details provided', async ({
   persistentContext,
--- a/tests/e2e/fixtures/base.ts
+++ b/tests/e2e/fixtures/base.ts
@@ -16,27 +16,22 @@ import { getLastCallArg } from '../helpers/common';
 import { openPopup, type Popup } from '../pages/popup';
 import type { DeepPartial, Storage } from '@/shared/types';
 
-type BaseScopeWorker = {
+type Fixtures = {
   persistentContext: BrowserContext;
   background: Background;
-  i18n: BrowserIntl;
-  /**
-   * IMPORTANT: This is created once per test file. Mutating/closing could
-   * impact other tests in same file.
-   */
   popup: Popup;
+  i18n: BrowserIntl;
+  page: Page;
 };
 
-export const test = base.extend<{ page: Page }, BaseScopeWorker>({
-  // Extensions only work with a persistent context.
-  // Ideally we wanted this fixture to be named "context", but it's already defined in default base context under the scope "test".
+export const test = base.extend<Fixtures>({
   persistentContext: [
-    async ({ browserName, channel }, use, workerInfo) => {
-      const context = await loadContext({ browserName, channel }, workerInfo);
+    async ({ browserName, channel }, use, testInfo) => {
+      const context = await loadContext({ browserName, channel }, testInfo);
       await use(context);
       await context.close();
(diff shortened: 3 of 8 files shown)


sidvishnoi (commented):
test-e2e

tests/e2e/connect.spec.ts
```diff
@@ -1,11 +1,6 @@
-/// <reference types="chrome"/>
 import { test, expect } from './fixtures/base';
 import { connectWallet, disconnectWallet } from './pages/popup';
 
-test.beforeEach(async ({ popup }) => {
-  await popup.reload();
-});
-
```
sidvishnoi, line 8: New popup is created for each test, so don't need reload.

tests/e2e/not-monetized.spec.ts
```diff
@@ -1,133 +1,124 @@
...
+    await expect(warning).toHaveText(msg.newTab);
   });
 
-  test('file:// URLs', async ({ page }) => {
-    await page.goto(pathToFileURL('.').href);
-    await expect(warning).toBeVisible();
-    await expect(warning).toHaveText(msg);
-  });
+  await test.step('shows not monetized on internal pages', async () => {
+    await using page = await newPage();
```
sidvishnoi, line 47: This along with `Symbol.asyncDispose` will automatically close the page at then end of this code block.

We also reload popup on each new page create - to simulate new tab closing the popup in browser.

sidvishnoi (commented):
test-e2e

tests/e2e/not-monetized.spec.ts
```diff
@@ -1,133 +1,124 @@
...
-  await expect(warning).toHaveText(msg);
-
-  await context.newPage().then((page) => page.bringToFront());
-  await expect(warning).toBeVisible();
-  await expect(warning).toHaveText(msg);
-});
-
-test('shows not monetized on internal pages', async ({
-  page,
+// We make extensive use of test.step in this test to reuse the same browser
```
raducristianpopa, line 7: State will be persisted between steps?

sidvishnoi: Yes, reusing same browser, which contains the state.

sidvishnoi: `test.step` is just for labeling blocks in a single test.

tests/e2e/fixtures/connected.ts
```diff
@@ -16,13 +16,13 @@ export const test = base.extend<{ page: Page }, { popup: Popup }>({
         amount: '10',
         recurring: false,
       });
-      await popup.reload({ waitUntil: 'networkidle' });
+      await popup.reload();
```
DarianM, line 19: reload still needed here? with new popup created for each test

sidvishnoi: A reload after connect, to simulate popup closing & reopening.

playwright.config.ts
```diff
@@ -9,15 +9,9 @@ if (!process.env.CI) {
...
-  // We don't want this set to true as that would make tests in each file to run
-  // in parallel, which will cause conflicts with the "global state". With this
-  // set to false and workers > 1, multiple test files can run in parallel, but
-  // tests within a file are run at one at a time. We make extensive use of
-  // worker-scope fixtures and beforeAll hooks to achieve best performance.
-  fullyParallel: false,
+  fullyParallel: true,
   forbidOnly: !!process.env.CI,
   retries: process.env.CI ? 2 : 0,
-  workers: process.env.CI ? 1 : undefined,
```
raducristianpopa, line 20: Can we have a number of max workers with `fullyParallel: true`?

tests/e2e/fixtures/base.ts
```diff
@@ -47,15 +42,15 @@ export const test = base.extend<{ page: Page }, BaseScopeWorker>({
       const background = await getBackground(browserName, context);
       await use(background);
     },
-    { scope: 'worker', timeout: 5_000 },
+    { scope: 'test', timeout: 5_000 },
```
raducristianpopa, line 45: Since these are probably needed only once per worker, why the switch to `test`?

playwright.config.ts
sidvishnoi: The default is half the available CPU cores. So, on GitHub it means `2` (there's 4 cores).
Can specify workers still, but better use as many as we can for speed. Can specify `--workers=N` from CLI if needed for one-off tests.

tests/e2e/fixtures/base.ts
sidvishnoi: Needed once per test. Otherwise it'll keeping using background across tests. Adding message to Darian for more context if that helps:

> A worker is something internal to playwright - a worker runs as long as it wants to. resources of a worker (see, worker scope) get cleaned only after worker ends.
> A worker can continue to run across multiple files.
> 
> `browser` (`persistentContext`) is independent of worker. it's up to us when to create it - if we create it in `worker` scope, it'll get reused across entire worker scope. if we create browser in `test` scope, it'll create a new one for each test.
> 
> (browser means: browser + extension + extension background service worker + popup)
>
> So, if you run with `--workers=2`, it'll create two workers for entirety of test. each worker (without dying) will run multiple tests across multiple files. a new worker will be created only if a test fails

DarianM (approved):
:shipit:

raducristianpopa (approved):

PR #886, opened 2025-02-02
fix(popup): remove amount from OTP msg

### LINKED ISSUE #859: Remove amount from payment confirmation message
- Partial payments confuse users (I tried sending 0.5, why extension says 0.25 was sent)
- Showing amount in case of successful payment doesn't add value much either.

Remove `amount` from payment confirmation message, keeping it limited to: "Thank you for your support".

DarianM:
## Context

Closes #859.

- Remove amount from payment confirmation message, keeping it limited to: "Thank you for your support".

### DIFF
--- a/src/shared/messages.ts
+++ b/src/shared/messages.ts
@@ -113,6 +113,7 @@ export interface PayWebsitePayload {
 
 export interface PayWebsiteResponse {
   type: 'full' | 'partial';
+  sentAmount: string;
 }
 
 export interface UpdateRateOfPayPayload {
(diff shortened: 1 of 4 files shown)


src/shared/messages.ts
```diff
@@ -113,8 +113,6 @@ export interface PayWebsitePayload {
 
 export interface PayWebsiteResponse {
   type: 'full' | 'partial';
-  sentAmount: string;
```
sidvishnoi, line 116: Might be worth keeping `sentAmount` in response, for better debugging/insights.

PR #893, opened 2025-02-04
fix: correctly handle reload, iframe nav and hashchange

### LINKED ISSUE #819: Site becomes not monetized on clicking 1password notification
1. Visit https://webmonetization.org/play/. 
2. Add a wallet address.
   - Site becomes monetized.
3. 1password shows notification iframe to (incorrectly) ask to save wallet address in vault. Click "close".
4. Site becomes non-monetized.

Trace:
```
monetization [PAYMENT SESSION] requestId=20b4547b-fcc2-4b02-9d0e-d5ac3a92b0dd; receiver=https://ilp-sandbox.chimoney.com/37294745

    Clearing timeout=11
debug @ paymentSession.ts:229
clearTimers @ paymentSession.ts:222
stop @ paymentSession.ts:208
clearSessionsByTabId @ tabState.ts:190
(anonymous) @ tabEvents.ts:94

tab-state Cleared 1 sessions for tab 2002829598.
```

Diagnosed to this part of code:
```ts
// tabEvents.ts 
  onUpdatedTab: CallbackTab<'onUpdated'> = (tabId, changeInfo, tab) => {
    /**
     * if loading and no url -> clear all sessions but not the overpaying state
     * if loading and url -> we need to check if state keys include this url.
     */
    if (changeInfo.status === 'loading') {
      const url = tab.url ? removeQueryParams(tab.url) : '';
      const clearOverpaying = this.tabState.shouldClearOverpaying(tabId, url);

      this.tabState.clearSessionsByTabId(tabId);
```

### LINKED ISSUE #892: Site not monetized hashchange
1. Visit a monetized site
2. Visit any link that changes URL hash.

`tabs.onUpdated` is fired, so overpaying state gets cleared as well as site becomes not monetized. Same root problem as https://github.com/interledger/web-monetization-extension/issues/819

sidvishnoi:
## Context

Fixes #819
Fixes #892

- In content script, fire `STOP_MONETIZATION` with intent=remove on `pagehide` event. This ensure we clear payment sessions for frame (top-level as well as iframe) when there's any navigation.
   - We use `pagehide` as it doesn't interfere with bf-cache.
- `onUpdatedTab` will now only clear overpaying state. It'll do so if URL of top-frame is changed.
   - It won't try clearing sessions on every `onUpdatedTab` (as there's no way to distinguish between page reload and iframe navigation in this event),
   - but for sanity, it'll clear sessions on navigation (in case content script didn't fire pagehide for some reason).
- We store tab URL in `TabState` now. This helps us see how the URL changed in `onUpdatedTab` - i.e. we can ignore hashChange for overpaying clear.
- There are some needless calls to "stop & clear" sessions, that can be avoided, but not worth the added complexity as wasted calls are essentially no-op (just a debug log "No active sessions" is printed)
   - A `STOP_MONETIZATION` message is sent on `visibilitychange` which fires right after `pagehide`
   - Tab close and a nav that changes URL also triggers all 3 events.
- Add more E2E tests, and refactor a common request interception helper for tests.


sidvishnoi (commented):
test-e2e

src/content/services/monetizationLinkManager.ts
```diff
@@ -97,6 +97,7 @@ export class MonetizationLinkManager extends EventEmitter {
     );
     this.window.removeEventListener('message', this.onWindowMessage);
     this.window.removeEventListener('focus', this.onFocus);
+    this.window.removeEventListener('pagehide', this.onPageHide);
```
DarianM, line 100: looks like `end` method is not called from anywhere. for the cleanup, we can use [disposer](		window.document.body.appendChild(iframe);
) on container directly

sidvishnoi: It's not yet called, as there's not a strong enough use case to call it. We would likely need it to support https://github.com/interledger/web-monetization-extension/issues/418 (I had a WIP branch, until it didn't work as we wanted).

src/background/services/tabEvents.ts
```diff
@@ -83,19 +83,19 @@ export class TabEvents {
...
-     * if loading and url -> we need to check if state keys include this url.
-     */
-    if (changeInfo.status === 'loading') {
-      const url = tab.url ? removeQueryParams(tab.url) : '';
-      const clearOverpaying = this.tabState.shouldClearOverpaying(tabId, url);
-
-      this.tabState.clearSessionsByTabId(tabId);
-      if (clearOverpaying) {
-        this.tabState.clearOverpayingByTabId(tabId);
+    if (changeInfo.status === 'loading' && changeInfo.url) {
```
DarianM, line 86: should we address the overpaying inside a iframe with an event as well? sometime

DarianM (approved):

PR #894, opened 2025-02-05
refactor(tests/e2e): rename `persistentContext` to `context`

sidvishnoi:
## Context

`persistentContext` as identifier was required when we were declaring it on worker scope. With test scope, we can directly call it `context` (overriding default `context`)
i.e. we cannot declare same identifier in different `scope`, but now we use `test` scope only (https://github.com/interledger/web-monetization-extension/pull/879)

Rename `persistentContext` to `context`


DarianM (commented):
test-e2e

DarianM (approved):

Issue #610, opened 2024-09-19
[TASK] Write end-to-end tests based on testing doc


sidvishnoi:
Using testing doc in https://github.com/interledger/web-monetization-extension/issues/290, write end-to-end tests.

sidvishnoi:
Closed with above linked PRs. There's more to test, but the cases in testing doc are addressed. Will follow up with another test story for future tests.

PR #904, opened 2025-02-06
test(e2e): add tests for out of funds & add funds

sidvishnoi:
## Context

Part of #610

### DIFF
--- a/tests/e2e/out-of-funds.spec.ts
+++ b/tests/e2e/out-of-funds.spec.ts
@@ -1,0 +1,300 @@
+import { test, expect } from './fixtures/base';
+import { getStorage } from './fixtures/helpers';
+import {
+  connectWallet,
+  disconnectWallet,
+  sendOneTimePayment,
+  setContinuousPayments,
+} from './pages/popup';
+import {
+  getContinueWaitTime,
+  getWalletInfoCached,
+  setupPlayground,
+} from './helpers/common';
+import { completeGrant, DEFAULT_CONTINUE_WAIT_MS } from './helpers/testWallet';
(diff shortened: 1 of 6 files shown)


sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

tests/e2e/out-of-funds.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+        await expect(monetizationCallback).toHaveBeenCalledTimes(2);
+      });
+
+      await test.step('trigger out-of-funds with a micro-payment', async () => {
+        await setContinuousPayments(popup, true);
+        await expect(background).toHaveStorage({
+          continuousPaymentsEnabled: true,
+        });
+        await popup.reload();
+        await page.bringToFront();
```
DarianM, line 133: loving the for loop test_cases. well done tests! 🎉 
it's not clear to me the necessity to do reload and brindtofront here, why doesn't the test pass without it?

sidvishnoi: I'll check. Reload might be needed, bring to front maybe not.

sidvishnoi: Oh, remembered now.
`popup.reload()` is needed as we're turning on continuous payments on settings screen. `reload()` will will bring popup back to home screen.
`page.bringToFront()` isn't required, but was added to ensure that the page is in focus so payments can go through. It isn't required apparently - I added it in case popup is in focus.

DarianM (approved):

Issue #847, opened 2025-01-17
Outgoing payment created before probing completed


sidvishnoi:
When continuous payments are "resumed", and probing was still active, our "locks" on adjustAmount get ignored. `PaymentSession#amount` is undefined by the time `payContinuous` was called due to `PaymentSession@resume` - which meant, `POST /outgoing-payments` request had missing `debitAmount.value` (the API showed it as "missing quoteId" error, due to schema validation returning first error it thought was the cause).

Trigger: test ran too fast, before probing was completed (300ms vs ~1.5s)

Found in extensive debugging session with @raducristianpopa (big thanks!) when figuring out why test in https://github.com/interledger/web-monetization-extension/pull/835 failed occasionally. cc @DarianM

PR #910, opened 2025-02-10
fix(bg/PaymentSession): ensure probing completed before start

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/847

- Replace probing abort logic to use `AbortSignal` instead of internal `probingId`
- Create a wrapper over `adjustAmount` to manage above abortion.
   - This wrapper preserves the last `adjustAmount` call's resulting promise (for the same `rate`), so we can call it again in `start` to ensure the promise is fulfilled before we actually create outgoing payment.
- Not the most elegant code, I know. Open to suggestions.

### DIFF
--- a/src/background/services/paymentSession.ts
+++ b/src/background/services/paymentSession.ts
@@ -104,7 +104,8 @@ export class PaymentSession {
       rate,
       this.#adjustAmountController.signal,
     );
-    await this.#adjustAmountPromise;
+
+    return this.#adjustAmountPromise;
   }
 
   private async _adjustAmount(
--- a/src/background/services/paymentSession.ts
+++ b/src/background/services/paymentSession.ts
@@ -64,7 +64,6 @@ export class PaymentSession {
   private incomingPaymentExpiresAt: number;
   private amount: string;
   private intervalInMs: number;
-  private probingId: number;
   private shouldRetryImmediately = false;
 
   private timeout: ReturnType<typeof setTimeout> | null = null;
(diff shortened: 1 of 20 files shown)


src/background/services/paymentSession.ts
```diff
@@ -264,8 +284,20 @@ export class PaymentSession {
...
     };
 
+    if (!this.rate) {
+      // this.rate is set when adjustAmount begins. this.amount is set only after first successful adjustAmount
+      throw new Error('Unexpected: adjustAmount not yet ready');
+    }
     if (this.canContinuePayment) {
       this.timeout = setTimeout(async () => {
+        if (!this.amount) {
+          await this.adjustAmount(this.rate);
```
sidvishnoi, line 294: This call here will reuse the existing probing promise instead of starting a new one.

src/background/services/paymentSession.ts
```diff
@@ -85,9 +84,33 @@ export class PaymentSession {
...
+          `Aborting existing probing for rate=${this.#adjustAmountLastRate}`,
+          'AbortError',
+        ),
+      );
+      this.#adjustAmountController = new AbortController();
+      this.#adjustAmountPromise = null;
+    }
+
+    this.#adjustAmountLastRate = rate;
+    this.#adjustAmountPromise ??= this._adjustAmount(
```
DarianM, line 103: nit: how about remove `??=` and assign the promise all the time. remove `this.#adjustAmountPromise = null;`

src/background/services/paymentSession.ts
```diff
@@ -85,9 +84,33 @@ export class PaymentSession {
...
+      rate,
+      this.#adjustAmountController.signal,
+    );
+    await this.#adjustAmountPromise;
```
DarianM, line 107: nit:     return this.#adjustAmountPromise;

src/background/services/paymentSession.ts
sidvishnoi: This will create a new promise, and not use existing `_adjustAmount` one (i.e. it'll restart probing from start everytime).

DarianM (commented):
test-e2e

DarianM (approved):

src/background/services/paymentSession.ts
```diff
@@ -85,9 +84,34 @@ export class PaymentSession {
     private message: MessageManager<BackgroundToContentMessage>,
   ) {}
 
+  #adjustAmountLastRate: AmountValue;
+  #adjustAmountController = new AbortController();
+  #adjustAmountPromise: null | Promise<void> = null;
   async adjustAmount(rate: AmountValue): Promise<void> {
```
sidvishnoi, line 90:
```suggestion
  adjustAmount(rate: AmountValue): Promise<void> {
```

sidvishnoi (commented):
Don't need async. Also, it'll prevent sending call to promise queue this way - should avoid some subtle timing issues

src/background/services/paymentSession.ts
DarianM: very good catch

DarianM: look at this [useAwait](https://biomejs.dev/linter/rules/use-await/#invalid); the rule would have caught this, let's use it

PR #914, opened 2025-02-12
test(e2e): add tests for multiple link tags - same/cross currency

sidvishnoi:
## Context

Part of https://github.com/interledger/web-monetization-extension/issues/610


src/background/utils.ts
```diff
@@ -114,6 +113,7 @@ export const redirectToWelcomeScreen = async (
   intent: InteractionIntent,
   errorCode?: ErrorCode,
 ): Promise<void> => {
+  const { OPEN_PAYMENTS_REDIRECT_URL } = await import('@/shared/defines');
```
sidvishnoi, line 116: Otherwise importing this file in tests says `OPEN_PAYMENTS_REDIRECT_URL` is not defined.

tests/e2e/multiple.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+
+// param amount is in bigint format (e.g. 1.2 at assetScale 2 would be 120)
+// 1 unit less by @interledger/pay and 1% transaction fee on test wallet.
+const getAmountAfterFee = (amount: number) =>
+  Math.round(Number((amount - 1) * 0.99));
+
+const orderById = <T extends { id: string }>(a: T, b: T) =>
+  a.id.localeCompare(b.id);
+
+test.describe('should monetized site with multiple wallet address', () => {
```
sidvishnoi, line 45: Will merge both tests and use a loop.

sidvishnoi:
Need to address https://github.com/interledger/web-monetization-extension/issues/915 first.

tests/e2e/multiple.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+      const outgoingPayments = outgoingPaymentCreatedCallback.calls.map(
+        ([{ receiveAmount, debitAmount }]) => ({
+          receiveAmount,
+          debitAmount,
+        }),
+      );
+
+      expect(
+        outgoingPayments,
+        'outgoing payments created with same split debit amount',
```
DarianM, line 155: i think this should be outside one-time payment step, inside it's own outgoing-payments check step

sidvishnoi: We're checking what amount the outgoing payments (for the one time payment) were created with here.

DarianM (approved):

PR #918, opened 2025-02-13
chore: additional lint rules

DarianM:

- Enabled the few ("off") lint rules
- Add `useAwait` [rule](https://biomejs.dev/linter/rules/use-await/)
- Disabled biome for `*.test.tsx` files (biome does not have disable linter per file/block for now; planned for v2.0.0.)

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -6,7 +6,7 @@
     "useIgnoreFile": true,
     "defaultBranch": "main"
   },
-  "files": { "ignoreUnknown": true, "ignore": [] },
+  "files": { "ignoreUnknown": true, "ignore": ["*.test.tsx"] },
   "formatter": {
     "enabled": true,
     "useEditorconfig": true,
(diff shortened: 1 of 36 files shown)


biome.jsonc
```diff
@@ -20,11 +20,12 @@
     "rules": {
       "recommended": true,
       "style": {
-        "noUselessElse": "off", // TODO: turn on, too many cases
-        "noNonNullAssertion": "off" // TODO: turn on, too many cases
+        "noUselessElse": "error",
+        "noNonNullAssertion": "error"
```
sidvishnoi, line 24: Let's not enable `noNonNullAssertion` as too many cases that we can handle better (it's ok to lint ignore sometimes)

biome.jsonc
```diff
@@ -20,11 +20,12 @@
...
       },
       "suspicious": {
-        "noConsole": "warn"
+        "noConsole": "warn",
+        "useAwait": "error"
```
sidvishnoi, line 28: Sometimes we can make diff smaller like this, so diff are easier to understand in future (like git blame).

```suggestion
        "useAwait": "error",
        "noConsole": "warn"
```

biome.jsonc
```diff
@@ -6,7 +6,7 @@
     "useIgnoreFile": true,
     "defaultBranch": "main"
   },
-  "files": { "ignoreUnknown": true, "ignore": [] },
+  "files": { "ignoreUnknown": true, "ignore": ["*.test.tsx"] },
```
sidvishnoi, line 9: Let's not do this if there are too many cases - better to ignore one lint rule than disable all for some files.

sidvishnoi:
At this stage, better create new PRs instead of trying fixing merge conflicts 😄

DarianM:
out of date, let's keep it as a future ref.
this needs to be broken into smaller rules at once.

closing this

Issue #921, opened 2025-02-13
At any time allow one single monetization link in iframes


DarianM:
- extension should monetize only the first link it finds inside first level monetization iframes head
- if multiple links are present and the first one gets removed/invalidated, start the monetization for next found link

suggestions to do this:
- creating a payment session for every link tag found in frame and start one when needed (i.e. when another gets invalidated) might be better than searching the head for the next valid link tag

part of #842

sidvishnoi:
Closing in favor of https://github.com/interledger/web-monetization-extension/issues/976.

PR #922, opened 2025-02-13
fix(MonetizationLinkManager): simplify search, get first link from iframe

### LINKED ISSUE #920: Simplify monetization link detection and handling
- instead of searching every added or removed node using the mutation observer, use the same observer to search the entire window / first level iframes document every time an element gets added
- make use of existing `getMonetizationLinkTags` logic that searches for link tangs inside documents and build around it

part of #842

DarianM:
Closes #920 .

- add mutation observer to first level frames
- removed onAddedNode/onRemovedNode
- use `onWholeDocumentObserved` to search every time a new node gets removed or added
_____________________
Current known limitation for when the first monetization link is removed, the second should be monetized, to be addressed in #921


DarianM (commented):
test-e2e

src/content/services/monetizationLinkManager.ts
```diff
@@ -370,27 +372,30 @@ export class MonetizationLinkManager extends EventEmitter {
...
-    const childListRecords = records.filter((e) => e.type === 'childList');
-    const removedNodes = childListRecords.flatMap((e) => [...e.removedNodes]);
-    const allRemovedLinkTags = removedNodes.map((node) =>
-      this.onRemovedNode(node),
-    );
-    const stopMonetizationPayload: StopMonetizationPayload = allRemovedLinkTags
-      .filter(isNotNull)
-      .flat();
-    await this.sendStopMonetization(stopMonetizationPayload);
+  private linkTags: Set<HTMLLinkElement> = new Set();
```
sidvishnoi, line 375: How is this different from `this.monetizationLinks`?

src/content/services/monetizationLinkManager.ts
```diff
@@ -161,27 +162,28 @@ export class MonetizationLinkManager extends EventEmitter {
...
         'head link[rel="monetization"]',
       );
-      return monetizationTag ? [monetizationTag] : [];
+      return new Set(monetizationTag ? [monetizationTag] : []);
     }
   }
 
   /** @throws never throws */
-  private async checkLink(link: HTMLLinkElement) {
+  private async getLinkDetails(link: HTMLLinkElement) {
```
sidvishnoi, line 186: Let's avoid renaming so we can get a clearer diff of what we're doing in this PR. Can do renaming as a follow up, so each merged commit only does one thing it aimed to do.

src/content/services/monetizationLinkManager.ts
```diff
@@ -40,7 +39,7 @@ export class MonetizationLinkManager extends EventEmitter {
     });
 
     this.documentObserver = new MutationObserver((records) =>
-      this.onWholeDocumentObserved(records),
+      this.onAddedOrRemovedNodesMutationObserverCallback(records),
```
sidvishnoi, line 42: I think can keep calling it `onWholeDocumentObserved` still. Just its inner working has changed.

src/content/services/monetizationLinkManager.ts
```diff
@@ -498,56 +498,40 @@ export class MonetizationLinkManager extends EventEmitter {
...
+    const promise = this.checkLink(link);
+    this.monetizationLinks.set(link, promise);
+
+    const response = await promise;
+    if (response) {
+      this.observeLinkAttrs(link);
     }
 
-    return null;
+    return promise;
```
sidvishnoi, line 523: I think we're making the wrong part async/promise-cache. The real async call in call stack that we want to _cache_ is `validateLink`, which does the message passing bit. Let's try caching there instead?

src/content/services/monetizationLinkManager.ts
```diff
@@ -507,61 +510,32 @@ export class MonetizationLinkManager {
...
-    }
-
-    return null;
-  }
-
   private async onAddedLink(
     link: HTMLLinkElement,
   ): Promise<StartMonetizationPayloadEntry | null> {
+    if (this.pendingValidationLinks.has(link)) {
+      return null;
```
sidvishnoi, line 517: Let's add a comment here why we do it.

src/content/services/monetizationLinkManager.ts
```diff
@@ -376,28 +374,28 @@ export class MonetizationLinkManager {
...
+        linkTagsNow,
+        new Set(this.monetizationLinks.keys()),
       );
-      const startMonetizationPayload = allAddedLinkTags
-        .filter(isNotNull)
-        .flat();
+      const linkTagEntries = [...tagsAdded].map((tag) => this.onAddedLink(tag));
+      for (const tag of tagsAdded) {
+        this.pendingValidationLinks.add(tag);
+      }
```
sidvishnoi, line 387: Can we do this within `onAddedLink`?

src/content/services/monetizationLinkManager.ts
```diff
@@ -376,28 +374,28 @@ export class MonetizationLinkManager {
   };
 
   private async onWholeDocumentObserved(records: MutationRecord[]) {
-    const { HTMLElement } = this.global;
+    if (this.isTopFrame || this.isFirstLevelFrame) {
```
sidvishnoi, line 377: Can we do an early return here?
```ts
if (!this.isTopFrame && !this.isFirstLevelFrame) return;
```

DarianM:
@sidvishnoi this should be reviewed and merged first
it does not have dynamic iframes tests enabled

src/content/services/monetizationLinkManager.ts
```diff
@@ -313,7 +309,9 @@ export class MonetizationLinkManager {
   ) {
     const payload: StopMonetizationPayload = [
       ...this.monetizationLinks.values(),
-    ].map(({ requestId }) => ({ requestId, intent }));
+    ]
+      .filter(isNotNull)
```
sidvishnoi, line 313: This is never null.
```suggestion
```

src/content/services/monetizationLinkManager.ts
```diff
@@ -411,6 +387,29 @@ export class MonetizationLinkManager {
         });
       }
     }
+
+    if (!this.isTopFrame && !this.isFirstLevelFrame) {
+      return;
+    }
```
sidvishnoi, line 393: Do we want to process attribute records when this is false? Should this move to top of this function?

src/content/services/monetizationLinkManager.ts
```diff
@@ -507,60 +518,38 @@ export class MonetizationLinkManager {
...
+
+    return walletAddress;
   }
 
   private onRemovedLink(link: HTMLLinkElement): StopMonetizationPayloadEntry {
     const details = this.monetizationLinks.get(link);
     if (!details) {
       throw new Error(
-        `Could not find details for monetization node ${link.outerHTML.slice(0, 200)}`,
+        `Could not find wallet address for monetization node ${link.outerHTML.slice(0, 200)}`,
```
sidvishnoi, line 550: This could confuse us. "find wallet address" could mean `GET` response, not the one that this is an unexpected error (a sanity check).
```suggestion
        `Unexpected: No details found for monetization node ${link.outerHTML.slice(0, 200)}`,
```

src/content/services/monetizationLinkManager.ts
```diff
@@ -133,20 +133,12 @@ export class MonetizationLinkManager {
     });
 
     const monetizationLinks = this.getMonetizationLinkTags();
-
-    for (const link of monetizationLinks) {
-      this.observeLinkAttrs(link);
-    }
```
sidvishnoi, line 139: Hmmm. This is now observing only valid link tags?

src/content/services/monetizationLinkManager.ts
DarianM: we're safe to move this to top. we shouldn't look for mutations outside main or first lvl iframes

src/content/services/monetizationLinkManager.ts
DarianM: changes added:

- additional link mutations do not recheck existing failed/invalid link monetization (and implicitly invalidating these over and over with each new mutation)
- now observing for attr changes in invalid tags

PR #923, opened 2025-02-14
chore(tests/e2e): make `env.d.ts` source of truth for env vars

sidvishnoi:
## Context

Instead of defining vars twice - once in markdown table, and again in env.d.ts, make env.d.ts source of truth. Defining in both places was getting out of hand.

- Update details in env.d.ts
- Create separate top-level interface `TestEnvVars`
- Remove unnecessary `TEST_WALLET_ORIGIN` variable (we can't really use localhost, as so many other dependent variables in test)


sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

DarianM (approved):

PR #927, opened 2025-02-19
fix: treat resume as start if sessions lost, trigger resume if needed

sidvishnoi:
## Context

Closes #841
Closes #871
More context: https://github.com/interledger/web-monetization-extension/issues/871#issuecomment-2663187885

## Changes proposed in this pull request

- Make ResumeMonetizationPayload to be same as StartMonetizationPayload
- When resume is called and background doesn't have sessions, use that payload to trigger a "start monetization". This addresses back-forward navigation sessions cleared issue.
- To address resume on sleep-clearing-all-sessions-in-background (here we don't have a payload we can trigger start from), pass `REQUEST_RESUME_MONETIZATION` message to content script to call resume in content, which will trigger the start.
- Trigger this resume on in `background.start()`.

## Testing

See https://github.com/interledger/web-monetization-extension/issues/871#issuecomment-2663187885 for how to reproduce. The bf-cache one is covered by E2E tests, but sleep one requires manual testing. I've tested following scenarios (with computer sent to sleep around 60-70 times 😆):
- Press Win+X U S to put computer to sleep (triggers window focus lose typically), Click start menu -> power -> sleep (doesn't trigger window focus lose typically)
- Keep monetized open when sending computer to sleep; keep window minimized when sending to sleep
- Keep a non-monetized tab active (window in view/minimized both cases)

PR #928, opened 2025-02-20
feat(scripts): allow remote logging during dev

sidvishnoi:
## Context

In https://github.com/interledger/web-monetization-extension/pull/927, the background will not terminate (even during sleep) if devtools console is open. So, to test this and similar issues, we allow enabling remote logging in dev.

- Add option to extend logger to emit logs (as JSON) to a remote server.
   - This is enabled by providing `LOG_SERVER=http://127.0.0.1:8000` env var in dev script
   - Production builds stay as is.
- Add a local Node server which emits (in pretty way) its POST body (which is the log)
   - Run it via `tsx scripts/log-server.ts`
- We can use this during E2E testing as we can't intercept console logs from extension's background worker in Playwright, if/when needed.

### DIFF
--- a/esbuild/dev.ts
+++ b/esbuild/dev.ts
@@ -30,6 +30,9 @@ export const getDevOptions = ({
       CONFIG_OPEN_PAYMENTS_REDIRECT_URL: JSON.stringify(
         'https://webmonetization.org/welcome',
       ),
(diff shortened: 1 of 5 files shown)


sidvishnoi:
Preview output in `scripts/log-server.ts`:

image

PR #929, opened 2025-02-20
test(MonetizationLinkManager): add unit tests for link detection

### LINKED ISSUE #925: Unit test coverage for monetization link detection
Entire `MonetizationLinkManager` class needs to be covered by unit tests before any new changes applied to detection.

part of https://github.com/interledger/web-monetization-extension/issues/842

DarianM:
## Context

Closes #925 .


sidvishnoi:
test-e2e

sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

src/content/__tests__/monetizationLinkManager.test.ts
```diff
@@ -91,6 +91,8 @@ describe('MonetizationLinkManager', () => {
 
   beforeEach(() => {
     jest.clearAllMocks();
+    // biome-ignore lint/complexity/noForEach: <explanation>
+    Object.values(msg).forEach((m) => m.mockReset());
```
sidvishnoi, line 95: `clearAllMocks` doesn't clear nested ones?

DarianM: `clearAllMocks` does not clear `mockResolvedValueOnce` queue. it looks like using mockResolvedValueOnce/mockReturnValue/mockImplementation, jest assumes you set up intentional behavior that should persist unless manually reset.

`clearAllMock` clears mock call history, but it does not remove implementations set via mockResolvedValueOnce.

`resetAllMocks` or `restoreAllMocks` don't work either

src/content/__tests__/monetizationLinkManager.test.ts
```diff
@@ -0,0 +1,569 @@
...
+
+      // should dispatch error event
+      expect(dispatchEventSpy).toHaveBeenCalledWith(new Event('error'));
+      expect(loggerMock.error).toHaveBeenCalled();
+
+      // should not start monetization
+      expect(msg.START_MONETIZATION).not.toHaveBeenCalledWith(
+        'START_MONETIZATION',
+        expect.any(Array),
+      );
```
sidvishnoi, line 250:
```suggestion
      expect(msg.START_MONETIZATION).not.toHaveBeenCalledWith(expect.any(Array));
```

src/content/__tests__/monetizationLinkManager.test.ts
```diff
@@ -0,0 +1,569 @@
...
+    });
+
+    test('should handle dynamically added monetization link in top frame', async () => {
+      const { document, documentReadyState } = createTestEnv({
+        head: '',
+      });
```
sidvishnoi, line 256:
```suggestion
      const { document, documentReadyState } = createTestEnv({});
```

src/content/__tests__/monetizationLinkManager.test.ts
```diff
@@ -0,0 +1,569 @@
...
+// for syntax highlighting
+const html = String.raw;
+
+const WALLET_ADDRESS: WalletAddress[] = [
+  {
+    authServer: 'https://auth.example.com',
+    publicName: 'Test Wallet USD',
+    assetCode: 'USD',
+    assetScale: 2,
+    id: '',
```
sidvishnoi, line 21: Let's add more real-ish values here. Then for wallet addresses, we can use `WALLET_ADDRESSES[0].id` instead of hard-coding values multiple times in tests.

src/content/__tests__/monetizationLinkManager.test.ts
```diff
@@ -0,0 +1,569 @@
+import { JSDOM } from 'jsdom';
+import { MonetizationLinkManager } from '@/content/services/monetizationLinkManager';
+import type {
+  ContentToBackgroundMessage,
+  MessageManager,
+  Response,
+} from '@/shared/messages';
+import { success, failure } from '@/shared/helpers';
```
sidvishnoi, line 8: Nit: Let's move this before the `type` imports. Helps grouping imports.

src/content/__tests__/monetizationLinkManager.test.ts
```diff
@@ -285,9 +334,255 @@ describe('MonetizationLinkManager', () => {
...
+        },
+        {
+          requestId: request2Id,
+          walletAddress: WALLET_ADDRESS[1],
+        },
+        {
+          requestId: request3Id,
+          walletAddress: WALLET_ADDRESS[2],
+        },
+      ]);
```
sidvishnoi, line 492: In this test, we want to see two `START_MONETIZATION` calls, one from each mutation callback.

src/content/__tests__/monetizationLinkManager.test.ts
```diff
@@ -0,0 +1,1236 @@
...
+    const window = dom.window;
+    const document = window.document;
+
+    return {
+      dom,
+      window,
+      document,
+      documentReadyState: jest.spyOn(document, 'readyState', 'get'),
+    };
+  }
```
sidvishnoi, line 98: Let's avoid repeating ourselves.

```suggestion
  function createTestEnv({
    head = '',
    body = '',
    interactive = false,
  }: { head?: string; body?: string; interactive?: boolean }) {
    const htm = html`<!DOCTYPE html><html><head>${head}</head><body>${body}</body></html>`;
    const dom = new JSDOM(htm, {
      runScripts: 'dangerously',
      pretendToBeVisual: true,
      resources: 'usable',
    });

    const window = dom.window;
    const document = window.document;

    const documentReadyState = jest.spyOn(document, 'readyState', 'get');
    if (interactive) {
      documentReadyState.mockReturnValue('interactive');
    }

    return {
      dom,
      window,
      document,
      documentReadyState,
    };
  }
```

sidvishnoi: Or maybe can pass `readyState = "interactive"` by default (as all current tests do that only), and then can override it manually later when needed. Like pass `{readyState:null}` to avoid setting it in `createTestEnv` and then later set it manually via `documentReadyState`.

src/content/__tests__/monetizationLinkManager.test.ts
```diff
@@ -0,0 +1,1236 @@
...
+  Response,
+} from '@/shared/messages';
+import type { Logger } from '@/shared/logger';
+import type { WalletAddress } from '@interledger/open-payments';
+import { success, failure } from '@/shared/helpers';
+
+// for syntax highlighting
+const html = String.raw;
+
+const WALLET_ADDRESS: WalletAddress[] = [
```
sidvishnoi, line 15: Can rename this to `WALLET_INFO` and create `WALLET_ADDRESS = WALLET_INFO.map(e => e.id);` then we don't have to use `WALLET_ADDRESS[0].id` everything in tests, and can do `WALLET_ADDRESS[0]` directly.

src/content/__tests__/monetizationLinkManager.test.ts
```diff
@@ -0,0 +1,1166 @@
...
+    resources: 'usable',
+  });
+
+  const window = dom.window;
+  const document = window.document;
+  const documentReadyState = jest.spyOn(document, 'readyState', 'get');
+
+  if (readyState === null) {
+    documentReadyState.mockReturnValue('interactive');
+  }
```
sidvishnoi, line 102:
```suggestion
  if (readyState) {
    documentReadyState.mockReturnValue(readyState);
  }
```

src/content/__tests__/monetizationLinkManager.test.ts
```diff
@@ -0,0 +1,1166 @@
...
+    logger: loggerMock,
+  });
+
+  return linkManager;
+}
+
+function createTestEnv({
+  head = '',
+  body = '',
+  readyState = null,
```
sidvishnoi, line 83:
```suggestion
  readyState = 'interactive',
```

src/content/services/monetizationLinkManager.ts
```diff
@@ -30,14 +32,19 @@ export class MonetizationLinkManager extends EventEmitter {
...
       document,
       logger,
       message,
+      window: global.window,
     });
+    const { MutationObserver, crypto, window } = this.global;
+
+    this.id = crypto.randomUUID();
+    this.isTopFrame = window === window.top;
+    this.isFirstLevelFrame = window.parent === window.top;
```
sidvishnoi, line 47: Moved these up otherwise mutation observer could be fired before these are assigned?! Saw happening in tests only.

src/content/__tests__/monetizationLinkManager.test.ts
```diff
@@ -478,33 +478,65 @@ describe('monetization in main frame', () => {
...
   });
 
-  test('handles rapid attribute changes on monetization link', async () => {
-    const { document } = createTestEnv({
-      head: html`<link rel="monetization" href="${WALLET_ADDRESS[0]}">`,
-    });
-    using linkManager = createMonetizationLinkManager(document);
+  test.failing(
+    'handles rapid attribute changes on monetization link',
+    async () => {
```
sidvishnoi, line 483: This should pass I believe.

sidvishnoi (commented):
test-e2e

src/content/__tests__/monetizationLinkManager.test.ts
sidvishnoi: In future then. I saw we don't "collect" quick attribute changes in MonetizationLinkManager

sidvishnoi (approved):
Thanks @DarianM for this great effort! Link manager will be much safer to modify & reason about now!
LGTM! 💯

Issue #844, opened 2025-01-17
Spinner keys showing in extension until connect process completed


sidvishnoi:
Spinner keeps showing (in connect wallet form) unless connect process is completed, error or tab closed.
Restarting the connect process needed (which happens on tab-close right now).

Things to be done:
- [x] Improve detection of "user getting lost" - like if user switched to some other URL in key addition tab or accept wallet consent, abort there, as user may lose way back (https://github.com/interledger/web-monetization-extension/pull/943)
- [ ] In ConnectWalletForm submit button, show current step (what are we waiting for user to do) along with the spinner (https://github.com/interledger/web-monetization-extension/pull/944)
- [x] If possible, when user opens popup again during connect process, bring user's attention to tab where they need to do things (as users lose track of tabs) (https://github.com/interledger/web-monetization-extension/pull/945)

Out of scope for this story (to come in next sprint):
- #948

<details>
<summary>Original issue details</summary>

Consider:
- show a "cancel" button in popup when in pending state, so user can manually abort and restart
- add a timeout on connect when no progress is being made (Q: does waiting for login identify as process?)
- tab-close is sure fire way to abort, so maybe popup can suggest closing tab that was opened by extension too?

</details>

DarianM:
Is the purpose of the `cancel` button or connection timeout helpful for errors beyond our control? like during the wallet provider connection process?

sidvishnoi:
It's better than user keep getting spinner, when there's no error but process is pending indefinitely. https://github.com/interledger/web-monetization-extension/issues/834 is also facing similar problem - it crashed (went into hang for some reason, but didn't throw) so our process didn't catch it. User would likely go back to extension popup at some time to see what's happening, and keep seeing spinner. So, a cancel button/timeout would be better for sure.

PR #943, opened 2025-02-27
fix(keyAutoAdd): abort if user navigates to some other website

sidvishnoi:
## Context

Part of https://github.com/interledger/web-monetization-extension/issues/844

- Add a helper `isUrlAllowed` to know if current URL (website) is acceptable for user to be on
   - We're adding some defense if user navigated to some other website (like read docs or just lost track) during connect process.
- Split content scripts to be per wallet, not per "content script" (i.e. register the same script with different ID for sandbox/prod wallets).
   - This helps us identify "matches" condition, to get allowed hosts for particular wallet (i.e. `isUrlAllowed`)
   - also, don't persist these scripts across sessions: so updates will register new scripts, without us needing to call updateScripts
- Add a small util `removeListeners` to remove all listeners at once. 
   - This fixes a bug I found during this PR that has existed before: if you closed tab before connect and then reopened-last-closed-tab, connect content script was started again, while background thought it had aborted.
- Doing this handling only in key addition process for now. Not on "waiting to accept grant" part.

## Demo

[Screencast from 2025-02-27 16-47-42.webm](https://github.com/user-attachments/assets/33be4437-ef5c-4291-88bf-34d73e26f808)

And a demo with GateHub, which has a separate subdomain for login than the actual wallet:

[Screencast from 2025-02-27 16-50-45.webm](https://github.com/user-attachments/assets/eafcfd48-8a81-432c-bbf7-3352dfedf851)


sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

DarianM (approved):

PR #946, opened 2025-02-27
refactor(background): better reuse tab helper; always reuse tabs

sidvishnoi:
## Context

https://github.com/interledger/web-monetization-extension/pull/945#discussion_r1973578895

## Changes proposed in this pull request

Make tab creation responsibility of caller, so we've tabId available sooner.
And this `tabId` corresponds to a tab that exists.

PR #944, opened 2025-02-27
fix(ConnectWalletForm): show current step along with spinner

sidvishnoi:
## Context

Part of https://github.com/interledger/web-monetization-extension/issues/844

## Changes proposed in this pull request

- Save current step title in `TransientState['connect']`
   - Also add little helpers for setting transient state
- Show current step title in connect form's submit button
   - Allow Button to show text along with spinner when `loading=true`

### Demo

[Screencast from 2025-02-28 17-18-30.webm](https://github.com/user-attachments/assets/ec12eee5-1054-4f44-831d-b24f4e1856bf)

PR #947, opened 2025-02-28
refactor(bg): split create & complete outgoing payment grant

sidvishnoi:
## Context

Seems counterintuitive, but:
- Helps with https://github.com/interledger/web-monetization-extension/pull/944 to show correct step (when initial create grant fails, and then we switch to adding key).
- Helps with https://github.com/interledger/web-monetization-extension/pull/946 as we open tab at right time, instead of before `createOutgoingPaymentGrant` (as create grant doesn't rely on browser tabs, only complete grant does).


sidvishnoi (commented):
test-e2e

DarianM (approved):

PR #949, opened 2025-03-01
fix(MonetizationLinkManager): monetize single link tag in iframes

### LINKED ISSUE #921: At any time allow one single monetization link in iframes

DarianM:
Closes #921 
______________________________________


sidvishnoi:
Closing as no longer needed as described in closed issue #921.

Issue #962, opened 2025-03-06
Considering replacing react-router with wouter


sidvishnoi:
Wouter: https://github.com/molefrog/wouter

Has almost same API as react-router-dom, and is much lighter, as there's no loader etc. features which we don't use.

PR #963, opened 2025-03-06
perf(popup,app): replace `react-router-dom` with `wouter`

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/962
Closes https://github.com/interledger/web-monetization-extension/issues/969

- replace `react-router-dom` with `wouter` in popup and app
- use hashRouter instead of memory router in `app`

```diff
-  dist/chrome/popup/popup.js                               321.8kb
+  dist/chrome/popup/popup.js                               253.6kb

-  dist/chrome/pages/app/app.js                             285.6kb
+  dist/chrome/pages/app/app.js                             216.4kb

- chrome.zip 996.46 KB
+ chrome.zip 950.82 KB
```


tests/e2e/pages/popup.ts
```diff
@@ -105,7 +105,7 @@ export async function setContinuousPayments(popup: Popup, enabled: boolean) {
 
 /** Whatever screen we're on in popup right now, take us to Home screen */
 export async function goToHome(popup: Popup) {
-  await popup.reload(); // reload is enough to reset state
+  await popup.goto(popup.url().replace(/#.*$/, ''));
```
sidvishnoi, line 108: TODO: Check if changing `location.hash` in `page.evaluate()` would be better.

src/pages/popup/pages/Settings.tsx
```diff
@@ -18,18 +18,18 @@ const isValidTabId = (id: string) => {
...
     'settings.tabId',
     TABS[0].id,
     { maxAge: 10 * 60 * 1000, validate: isValidTabId },
   );
-  const tabIdFromState =
+  const tabIdFromState = null; /*
     location.state?.tabId && isValidTabId(location.state?.tabId)
       ? (location.state.tabId as string)
-      : null;
+      : null; */
```
sidvishnoi, line 32: This feature allowed us using `history.state` to move to specific tab. Will check why it doesn't work with `wouter`. Alternatively, might create separate pages for each tab, so we use can reuse router instead of a separate Tab plugin, and can switch to tabs via URLs (helpful during tests too).

src/pages/app/App.tsx
```diff
@@ -1,37 +1,39 @@
...
-    lazy: () => import('./pages/PostInstall'),
-  },
-] satisfies RouteObject[];
+const Routes = () => {
+  return (
+    <Suspense>
+      <Switch>
+        <Route
+          path={ROUTES.DEFAULT}
+          component={lazy(() => import('./pages/PostInstall'))}
```
sidvishnoi, line 21: Suspense is needed due to `lazy`. wouter doesn't native `lazy` like `react-router-dom`.

sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

sidvishnoi (commented):
test-e2e

src/pages/popup/components/Settings/WalletInformation.tsx
```diff
@@ -28,6 +31,7 @@ export const WalletInformation = ({
           ev.preventDefault();
           setIsSubmitting(true);
           await message.send('DISCONNECT_WALLET');
+          navigate(ROUTES_PATH.HOME);
           window.location.reload();
```
DarianM, line 35: is `window.location.reload()` still needed here? seen it in a couple of places after navigate

src/pages/popup/pages/Settings.tsx
DarianM: where / when would we need this feature?

src/pages/popup/components/Settings/WalletInformation.tsx
sidvishnoi: Often the `message.send` doesn't result in a state change on client (no `dispatch` called in response). Ideally, we'd `dispatch('SET_CONNECT_STATE', boolean); navigate()`, but that's more work. So, we go the lazy way and get all data fresh with a reload.

DarianM (approved):

src/pages/popup/pages/Settings.tsx
sidvishnoi: It existed once, when follow text was clickable and sent user directly to rate of pay tab, instead of settings home. If we don't have tabs, we can send to right page via URL anyway.

image

Issue #976, opened 2025-03-13
Continuous payment streams overhaul


sidvishnoi:
This is relevant when the page has more than one wallet address.

Current state:
- We pay all payment sessions together on page load, and then compute a wait time for next payments.
   - With many wallet addresses on page, a website can take a lot of money out of user's wallet, as we pay a minimum 2c (when USD-USD wallets) on each session, and we pay all the sessions. We don't skip any wallet addresses or use staggered/alternate payments to respect the payment rate set by user. There's no max limit on payment sessions as of now.
   - e.g. If user visits 2 websites with 100 wallet addresses, user loses $2 on just their page load (well takes around 1-4s to setup all sessions).
- For iframes, we grab only the first link tag found on page load time, to prevent abuse by iframes adding many link tags.
   - If link tag is added dynamically, it's ignored. If first link tag is removed or new one prepended, it's not monetized.
- This payment model is bit rudimentary (overpays, doesn't respect rate of pay correctly), but still complex, specially when it comes to iframes.

Target:
- We want to monetize all wallet addresses found on a page and in all iframes.
- We want to respect user's set payment rate more strongly and prevent overpaying any website.
- We don't want iframes to add 100 links and take from main website's share.

Tasks:
- [ ] Monetize all wallet addresses found on a page and in all iframes.
- [ ] Add a concept of payment queues which allow sending payments for different sessions in a staged/staggered manner (queue, one after other), to prevent overpaying on page load.
- [ ] Split amount such that:
   - For the host website, all wallet addresses get equally split amount from total amount that will be sent according to rate of pay.
   - Each iframe is treated as a single payment stream (gets treated as if a single wallet address on main page), and all link tags in iframe get an amount split from iframe's entitled amount and use staggered payments within itself.

> [!Note]
> This section needs more planning and reasoning first, but here's a rough idea:
> 
> e.g. the website (W) has 4 link tags (W1...W4), there's 1 iframe (IA) with 2 link tags (IA1, IA2), and 1 iframe (IB) with 1 link tag (IA1).
> - `amount` (as per rate of pay) will be split into 6 parts
> - Website (W) gets 4/6 of the amount: `1/6(W1) + 1/6(W2) + 1/6(W3) + 1/6(W4)`
> - Iframe 1 (IA) gets 1/6 of the amount: `1/12(IA1) + 1/12(IA2)`
> - Iframe 2 (IB) gets 1/6 of the amount: `1/6(IB1)`
[...]

PR #1046, opened 2025-05-01
perf(bg/monetization): reuse `PaymentSession` when link added

sidvishnoi:
## Context

The goal is to efficiently find the `minSendAmount` for a `PaymentSession`.

Reuse PaymentSession instead of starting a new one.

<details>
<summary>Original PR summary</summary>

- In the first commit (https://github.com/interledger/web-monetization-extension/pull/1046/commits/e4a8ae51ec998fe3d2829198d429d58e2c4b4088), it won't have the desired effect, as we have an `AbortSignal` that makes the probing restart.
- In the next commit (https://github.com/interledger/web-monetization-extension/pull/1046/commits/c5de59ff43e6572a62637b7099096d0536062a7a), we replace `adjustAmount(rate)` with `findMinSendAmount()` (which is independent of rate). Extracted into [1048](https://github.com/interledger/web-monetization-extension/pull/1048).
- When a new link tag is added, we want to keep using older `minSendAmount`, even if the `adjustAmount` may change due to splits. `adjustAmount` resulting rate is >= `minSendAmount`, so it makes sense to find `minSendAmount` first, and not find it ever again.
   - Currency fluctuations can change this `minSendAmount` in future, but then we'd need to probe/`findMinSendAmount` again (existing bug, will fix some time later).

~Also updated the constructor type for `PaymentSession` a bit.~ Moved to [#1047](https://github.com/interledger/web-monetization-extension/pull/1047). I was thinking of adding a static `PaymentSession.from()`, but we don't need it right now. The new constructor format will help there as well. 

</details>

---

Before (see quotes requests increasing as link tags added to page, including for existing tags):

https://github.com/user-attachments/assets/ba9aeb97-c9ed-4722-ac5e-ed0b35f98cca

After (see quotes requests for link tags already added are skipped; fewer requests): 

https://github.com/user-attachments/assets/167e95a4-2e02-4127-b776-833e7a076e28

### DIFF
--- a/src/background/services/monetization.ts
+++ b/src/background/services/monetization.ts
@@ -561,7 +561,7 @@ export class MonetizationService {
       return true;
     } catch (err) {
       if (err.name === 'AbortError') {
-        this.logger.debug('adjustAmount aborted due to new call');
+        this.logger.debug('setRate aborted due to new call', { err });
         return false;
       } else {
         throw err;
--- a/src/background/services/monetization.ts
+++ b/src/background/services/monetization.ts
@@ -557,11 +557,11 @@ export class MonetizationService {
     rate: AmountValue,
   ): Promise<boolean> {
     try {
-      await Promise.all(sessions.map((session) => session.setRate(rate)));
+      await Promise.all(sessions.map((session) => session.adjustAmount(rate)));
       return true;
     } catch (err) {
       if (err.name === 'AbortError') {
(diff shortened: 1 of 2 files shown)


sidvishnoi (commented):
test-e2e
**Edit:** Unrelated tests fail (wallet budget), so all good.

DarianM (approved):

PR #1048, opened 2025-05-02
fix(PaymentSession): find `minSendAmount` independent of rate

sidvishnoi:
## Context

Extracted from https://github.com/interledger/web-monetization-extension/pull/1046

- When a new link tag is added, we want to keep using older `minSendAmount`, even if the `adjustAmount` may change due to splits. `adjustAmount` resulting rate is >= `minSendAmount`, so it makes sense to find `minSendAmount` first, and not find it ever again.

Known existing issues (discovered while working this out):
- Currency fluctuations can change this `minSendAmount` in future. In that case, we'd need to probe/`findMinSendAmount` again.
- Exponential probing is likely not the best approach. (Raw thoughts follow) I wonder if we can use currency conversion and then binary search or two-way search some values (specially relevant when sending from e.g. MXN to USD: with exp probing, we go 1, 2, 4, 8... with some two way search, we can go like: 21 (ceil exchange rate), 22, 19)
- Can we only pay in multiples of `minSendAmount`? Need to check how OP/Rafiki behaves in this scenario. For example, when sending ZAR to USD (20 units vs 1), minSendAmount can be 20. Rate of pay can say we've to pay 30, but we can pay either 20 or 40 (1 or 2 units only). OP allows this, but is this optimum distribution?

### DIFF
--- a/src/background/services/paymentSession.ts
+++ b/src/background/services/paymentSession.ts
@@ -84,56 +84,46 @@ export class PaymentSession {
     Object.assign(this, this.deps);
   }
 
-  #adjustAmountLastRate: AmountValue;
-  #adjustAmountController = new AbortController();
-  #adjustAmountPromise: null | Promise<void> = null;
-  adjustAmount(rate: AmountValue): Promise<void> {
-    if (this.#adjustAmountLastRate && rate !== this.#adjustAmountLastRate) {
-      this.#adjustAmountController.abort(
-        new DOMException(
-          `Aborting existing probing for rate=${this.#adjustAmountLastRate}`,
-          'AbortError',
-        ),
-      );
-      this.#adjustAmountController = new AbortController();
-      this.#adjustAmountPromise = null;
-    }
+  // We keep setting #minSendAmount to non-zero values as we probe. Instead of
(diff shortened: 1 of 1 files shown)


src/background/services/paymentSession.ts
```diff
@@ -84,56 +84,46 @@ export class PaymentSession {
...
-
-    return this.#adjustAmountPromise;
+  // We keep setting #minSendAmount to non-zero values as we probe. Instead of
+  // checking #minSendAmount > 0, use this boolean to know if probing completed.
+  #minSendAmountFound = false;
+  #minSendAmount = 0n;
+  #minSendAmountPromise: ReturnType<typeof this._findMinSendAmount>;
+
+  findMinSendAmount(): Promise<void> {
+    this.#minSendAmountPromise ??= this._findMinSendAmount();
```
sidvishnoi, line 94: This ensures we call `_findMinSendAmount()` only once per `PaymentSession` instance.

DarianM (commented):
test-e2e

sidvishnoi:
Aside: I really need to figure out `page.evaluate: Execution context was destroyed` errors in E2E!

DarianM (approved):

PR #1056, opened 2025-05-15
perf(bg/PaymentSession): fewer quote requests using exchange rate

sidvishnoi:
## Context

When sending cross currency payments (specially from higher to lower exchange rate, e.g. MXN to USD (0.19MXN = 0.01USD)), make use of exchange rate and use it as the starting point for probing.

When sending from MXN to USD, there's now a single (1) quote instead of 6.
This won't be needed when we get with https://github.com/interledger/rafiki/pull/3411 available across wallets.

This also gives us a more accurate minSendAmount, as otherwise with current exponential probing, we may end up trying to send 1,2,4,8,16,32 units, instead of something in between (19 here).

The exchange rates are cached for 15min as we don't want to fetch them for each session.

### DIFF
--- a/src/background/services/paymentSession.ts
+++ b/src/background/services/paymentSession.ts
@@ -15,7 +15,13 @@ import {
   isOutOfBalanceError,
   isTokenExpiredError,
 } from '@/background/services/openPayments';
-import { bigIntMax, convert, getNextSendableAmount } from '@/background/utils';
+import {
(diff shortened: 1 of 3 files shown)


sidvishnoi (commented):
test-e2e
Edit: Unrelated tests fails only in Edge, so all good.

DarianM (approved):

PR #1066, opened 2025-06-06
feat!: overhaul continuous payments logic

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/976

### Overview

Major change: This affects our handling of continuous payments when there are more than 1 monetization link tags on page.

- Loop through `PaymentSession`s in a staggered manner while making continuous payments. 
   - Earlier, on page load, we used to pay all sessions at once, then adding a longer wait on next payments. Now, we pay one session at a time, and give other sessions smaller wait than before.
   - This respects user's rate of pay and prevents abuse by websites as user won't end up paying 100 units on page load if there are 100 link tags on page.
   - `MonetizationEvent` is fired more often than earlier, giving more continuous payments feel.
- Payments are made only in multiple of `PaymentSession.minSendAmount`, making payments fairer to sender and ASEs (no rounding loss/gains due to currency conversion).
- There's a `MIN_PAYMENT_WAIT` of 2s, to reduce load of OP servers.

### Changes

- Move payment timing and amount logic from individual payment sessions (per link tag) to `PaymentManager` (per tab).
   - There's no `PaymentSession.start()`, `PaymentSession.stop()` etc. anymore
   - Some changes in math and timing of payment amounts
      - Added concept of `PaymentManager.pendingAmount` which increments as per units-duration defined by user's rate of pay.
      - When `pendingAmount` reaches next payable session's `minSendAmount` (or higher), pay appropriate amount to that session and subtract paid amount from `pendingAmount`, and move onto next payable session.
      - Essentially, `pendingAmount` is like a bucket, and there's a FIFO queue of sessions waiting to empty it as per their needs.
   - Payment streams looping flow and demo: see below in details
- Made overpaying logic state and logic simpler
   - overpaying wait etc. no longer depends on walletAddress, but only of tab's URL.
   - walletAddress info is still used to emit the older monetization event
- Renamed `PaymentSession.pay(amount)` to `PaymentSession.payOneTime(amount)` for clarity.
- Made `PaymentSession.payContinuous()` public (and renamed to `.pay(amount)`, and added a helper (`payWithRetry(amount)`) that retries this payment on certain errors.
- Made concept of "stop" and "pause" streaming more distinguishable.
- Made use of a pausable Timer class to keep timings sane.

### Follow-up

- Some parts of code are not as efficient as they could be. e.g. we loop through sessions a lot, even when they may not be changed since last call (finding sessions to pay, size of sessions). Some things can be improved while still preserving code readability - like maintain a count of size on add/remove session, instead of iterating all frames and adding size of each. 
- Update E2E tests and add more unit tests.

---

<details>
<summary><h3>Payment streams looping flow and demo</h3></summary>

- First go through all payable link tags on main website, one by one
- Then pay first link in first iframe, then first link in second iframe
- Then again go through all payable link tags on main website
- Then pay second link of first iframe, then second link (if doesn't exist, then first again) of second iframe
- Then again go through all payable link tags on main website
- Then again pay first link in first iframe, then first link in second iframe, and so on

In this demo, there are 2 iframes. Each frame (including main site) has 2 link tags in same order: MXN and then ZAR.

Most importantly, payment rate is now respected! As an example, our rate of pay says 1c every 2s.
Earlier: when we had 6 link tags, we’d emit 6 monetization events on page load. Spending 6c on page load. Then we’d wait 12s for next payment, and pay 6c again.
Now: We’d emit one monetization event every 2s!

Of course, this is the case when sending wallet and receiving wallet have 1c = 1c exchange rate. When sending from MXN to USD, even though sender’s rate is 0.01MXN every 2s, USD wallet cannot receive an amount this small. It needs at least 0.20MXN to receive 0.01USD. So, we’ll wait 40s (such that amount is now 0.20MXN) for the next payment!

Video: https://drive.google.com/file/d/1LtcdAJmMnYuT6wCNCFLp4aGkJY9U5g9j/view?usp=sharing

</summary>

### DIFF
--- a/src/background/services/background.ts
+++ b/src/background/services/background.ts
@@ -22,6 +22,7 @@ export class Background {
   private monetizationService: Cradle['monetizationService'];
   private storage: Cradle['storage'];
   private logger: Cradle['logger'];
+  private tabState: Cradle['tabState'];
   private tabEvents: Cradle['tabEvents'];
   private windowState: Cradle['windowState'];
   private sendToPopup: Cradle['sendToPopup'];
@@ -35,6 +36,7 @@ export class Background {
     monetizationService,
     storage,
     logger,
+    tabState,
     tabEvents,
     windowState,
     sendToPopup,
@@ -49,6 +51,7 @@ export class Background {
       storage,
       sendToPopup,
       sendToApp,
+      tabState,
       tabEvents,
       windowState,
       logger,
@@ -259,6 +262,7 @@ export class Background {
 
             case 'DISCONNECT_WALLET':
               await this.walletService.disconnectWallet();
+              this.tabState.clearAllState('disconnect');
               await this.browser.alarms.clear(ALARM_RESET_OUT_OF_FUNDS);
               await this.updateVisualIndicatorsForCurrentTab();
               this.sendToPopup.send('SET_STATE', { state: {}, prevState: {} });
--- a/src/background/services/tabState.ts
+++ b/src/background/services/tabState.ts
@@ -144,6 +144,13 @@ export class TabState {
     }
   }
 
+  clearAllState(reason: string) {
+    this.logger.info('Clear all state', { reason });
+    this.state.clear();
+    this.currentIcon.clear();
+    this.paymentManagers.destroyAll(reason);
+  }
+
   clearSessionsByTabId(tabId: TabId) {
     this.currentIcon.delete(tabId);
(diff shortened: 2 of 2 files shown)


src/shared/helpers/misc.ts
sidvishnoi, line 1: I'll move this into a separate PR. It was caching errors as well (still not fixed).

Issue #977, opened 2025-03-13
[BUG] Default Amoutn should be 5 USD - or the equivalent on local


ioana-chiorean:
### Steps to reproduce

At this point we offer as default value 5 of local currency. 

### Expected result

We should offer the converted amount from 5 USD in the local amount 
(mentioned by @AlexLakatos )

### Actual result

At this point we offer as default value 5 of local currency. 

### Operating system

macOS

### Operating system version

NA

### Browsers

Firefox

### Browser version

NA

### Extension version

latest

sidvishnoi:
A plain currency conversion would give odd numbers (like 5 USD is 7.24 CAD) and won't take purchasing power parity of different currencies. 

Might be high time to look into that crowd sourced list of recommended budgets and payment rates we discussed in one of open payments call. It might help with native browser implementations as well.

In either approach, providing prettier numbers would be a good idea.

ioana-chiorean:
Decision WM Call 25/03 - estimate/round it up to the closest nice integer. 
Maintain a monthly updates estimation list.

jeremiahlee:
(Copying over my comment from Slack)

There are established purchasing power parity (PPP) indices that can inform a better default than converting $5 to a local currency. [The Big Mac Index](https://en.wikipedia.org/wiki/Big_Mac_Index) is available as [open source](https://github.com/TheEconomist/big-mac-data). I recommend setting the monthly budget to the price of a Big Mac, rounded to the nearest integer and rate of that amount/60 (60 = 2 hours a day, 30 day month).

PR #1073, opened 2025-06-10
feat: use per-currency default budget and rate of pay

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/977

### DIFF
--- a/src/background/config.ts
+++ b/src/background/config.ts
@@ -7,6 +7,8 @@ export const MAX_RATE_OF_PAY = '100';
 
 export const EXCHANGE_RATES_URL =
   'https://telemetry-exchange-rates.s3.amazonaws.com/exchange-rates-usd.json';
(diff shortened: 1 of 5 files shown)


kjmitchelljr (approved):

src/background/config.ts
```diff
@@ -9,6 +10,8 @@ export const MIN_PAYMENT_WAIT = 2000;
 
 export const EXCHANGE_RATES_URL =
   'https://telemetry-exchange-rates.s3.amazonaws.com/exchange-rates-usd.json';
+export const BUDGET_RECOMMENDATIONS_URL =
+  'https://interledger.github.io/web-monetization-budget-suggestions/v1/data.json';
```
DarianM, line 14: why do we also have a `max` value for each currency ?

PR #1096, opened 2025-06-13
refactor(bg): pass constructible classes in deps for testability

sidvishnoi:
## Context

Had some errors when trying to test `PaymentManager` class - as jest didn't quite like the imports - open-payments SDK caused issues with testing.

- Add PaymentManager and PaymentSession to awilix Cradle
- From Monetization, call the PaymentManager from dependency, instead of importing.
- From PaymentManager, call the PaymentSession from dependency, instead of importing.

Makes things easier to mock in tests now.


src/background/services/paymentManager.ts
```diff
@@ -202,6 +201,7 @@ export class PaymentManager {
     payableSessions: PaymentSession[],
     signal?: AbortSignal,
   ) {
+    const { isOutOfBalanceError } = await import('./openPayments');
```
sidvishnoi, line 204: Direct import was having issue as well with testing - similar to some other modules.

DarianM (approved):

Issue #905, opened 2025-02-07
[BUG] Max payment amount isn't respected


sidvishnoi:
### Steps to reproduce

1. Disable continuous payments
2. Go to a page with 2 links of same asset scale e.g. [playground and add two monetization links](https://webmonetization.org/play/?wa=https://ilp.interledger-test.dev/sid&wa=https://ilp.interledger-test.dev/sid-mxn-1)
3. Send payment of $0.03

### Expected result

An amount total less than or equal to $0.03 be sent, spread across both payment pointers. If amount can't be split, the last wallet address(es) should be skipped.

### Actual result

Total sent amount was $0.04 (as min amount that can be sent is $0.02).

### Operating system

Windows, macOS, Linux

### Browsers

Chrome, Microsoft Edge, Firefox

### Extension version

v0.5.0.1 (problem in all versions)

sidvishnoi:
Similar problem likely exists in continuous payments when there are multiple payment pointers. Will investigate. Might need to send payments in staggered manner there, instead of all together.

PR #1098, opened 2025-06-13
fix!(paymentManager): improve one-time payment distribution

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/905


src/background/services/paymentManager.ts
```diff
@@ -178,21 +178,62 @@ export class PaymentManager {
...
+          walletAddress: session.receiver.id,
+          amount: amount.toString(),
+          ...(outgoingPayment
+            ? {
+                debitAmount: outgoingPayment.debitAmount.value,
+                sentAmount: outgoingPayment.sentAmount.value,
+              }
+            : {}),
+        };
+      }),
```
sidvishnoi, line 236: It's not used right now, but can be useful with tracking (to know how well our algorithm works in practice).

DarianM (approved):

PR #1085, opened 2025-06-11
fix: remove `minRateOfPay`, make `WalletInfo.url` required

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/1084
Helps with https://github.com/interledger/web-monetization-extension/issues/1082

## Changes proposed in this pull request

- Remove `minRateOfPay` from code, storage and UI.
- In Settings > Rate of Pay, set min to 1 unit
- As we're doing a storage structure migration, make `WalletInfo.url` as required. Make it same as `WalletAddress.id` when not available. Helps with [#1082](https://github.com/interledger/web-monetization-extension/issues/1082).
- Return `WalletInfo` from `getWalletInformation` (instead of `WalletAddress`)

PR #1099, opened 2025-06-17
fix(bg,form): dynamically change min send amount in pay-form

### LINKED ISSUE #1071: min send amount in pay-form is hard coded at 0.02
When sending one-time payments, the minimum sendable amount is hard-coded at 2 units, regardless of connected wallet and wallet addresses on the website.

When sending from MXN to USD, the min sendable amount should be around 0.19MXN. Currently the transactions fail, but the user doesn't know why they failed ("We were unable to process this transaction. Please try again!").

This will go hand-in-hand with https://github.com/interledger/web-monetization-extension/issues/905.

We should set the min in the one-time-pay form to be dependent on current website's `minSendAmount` value. Relevant files to be updated:
- https://github.com/interledger/web-monetization-extension/blob/826cbb842e509f6d31425ef32be1dbc5bd7551ac/src/background/services/monetization.ts#L414
- https://github.com/interledger/web-monetization-extension/blob/826cbb842e509f6d31425ef32be1dbc5bd7551ac/src/background/services/paymentManager.ts#L135 
- https://github.com/interledger/web-monetization-extension/blob/826cbb842e509f6d31425ef32be1dbc5bd7551ac/src/pages/popup/components/PayWebsiteForm.tsx#L103-L106

The `PaymentManager` would need a `get minSendAmount(): bigint` which returns `minSendAmount` across all payable sessions

kjmitchelljr:
## Context

Closes #1071

Removes the hard coded value of $0.02 in the Pay Website Form for the minimum amount.

The change should reflect the currency of the wallet for the minimum payment to send.

### DIFF
--- a/src/background/services/monetization.ts
+++ b/src/background/services/monetization.ts
@@ -434,11 +434,22 @@ export class MonetizationService {
 
     const { oneTimeGrant, recurringGrant, ...dataFromStorage } = storedData;
 
+    // #1071
+    let minSendAmount = '0';
+      if (tab.id) {
+        // If we have a payment manager for this tab, get the min send amount
+        const paymentManager = this.tabState.paymentManagers.get(tab.id);
+        if (paymentManager) {
+          minSendAmount = paymentManager.minSendAmount.toString();
+        }
+      }
+
     return {
       ...dataFromStorage,
       balance: balance.total.toString(),
       tab: this.tabState.getPopupTabData(tab),
       transientState: this.storage.getPopupTransientState(),
+      minSendAmount,
       grants: {
         oneTime: oneTimeGrant?.amount,
         recurring: recurringGrant?.amount,
--- a/src/background/services/paymentManager.ts
+++ b/src/background/services/paymentManager.ts
@@ -150,6 +150,18 @@ export class PaymentManager {
     return this.sessions.filter((s) => s.isUsable);
   }
 
+  // bugfix #1071
+  get minSendAmount(): bigint {
+    // Check to see if there are any payable sessions
+    if (!this.payableSessions.length) return 0n;
+
+    // NOTE: How is a payable session defined, i.e. are could the minSendAmount be different currencies?
+    return this.payableSessions.reduce(
+      (max, session) => session.minSendAmount > max ? session.minSendAmount : max,
+      0n
+    );
+  }
+
   private createStreamIfNotExists(frameId: FrameId) {
     let stream = this.streams.get(frameId);
     if (!stream) {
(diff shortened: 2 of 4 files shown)


src/background/services/monetization.ts
```diff
@@ -434,11 +434,22 @@ export class MonetizationService {
 
     const { oneTimeGrant, recurringGrant, ...dataFromStorage } = storedData;
 
+    // #1071
```
sidvishnoi, line 437: Nit: don't need these comments

src/background/services/monetization.ts
```diff
@@ -434,11 +434,22 @@ export class MonetizationService {
...
+    let minSendAmount = '0';
+      if (tab.id) {
+        // If we have a payment manager for this tab, get the min send amount
+        const paymentManager = this.tabState.paymentManagers.get(tab.id);
+        if (paymentManager) {
+          minSendAmount = paymentManager.minSendAmount.toString();
+        }
+      }
```
sidvishnoi, line 445: Can simplify it to 

```ts
const minSendAmount = this.paymentManagers.get(tab.id)?.minSendAmount.toString() ?? '0';
```

src/background/services/monetization.ts
```diff
@@ -434,11 +434,22 @@ export class MonetizationService {
...
+
     return {
       ...dataFromStorage,
       balance: balance.total.toString(),
       tab: this.tabState.getPopupTabData(tab),
       transientState: this.storage.getPopupTransientState(),
+      minSendAmount,
```
sidvishnoi, line 452: Given this data is specific current tab, can you move this to `tab` key's value? Tab info changes often.

src/background/services/paymentManager.ts
```diff
@@ -150,6 +150,18 @@ export class PaymentManager {
     return this.sessions.filter((s) => s.isUsable);
   }
 
+  // bugfix #1071
+  get minSendAmount(): bigint {
+    // Check to see if there are any payable sessions
+    if (!this.payableSessions.length) return 0n;
```
sidvishnoi, line 156: Don't need this given `reduce` will handle the empty array case as well.

src/background/services/paymentManager.ts
```diff
@@ -150,6 +150,18 @@ export class PaymentManager {
...
+
+    // NOTE: How is a payable session defined, i.e. are could the minSendAmount be different currencies?
```
sidvishnoi, line 158: payable session is defined here:

https://github.com/interledger/web-monetization-extension/blob/893bf23e5b777d0c8cee01b31d91eef4429d3aba/src/background/services/paymentManager.ts#L149-L151

sidvishnoi (changes requested):

sidvishnoi:
> I noticed in storage there does exist `minRateOfPay`

It doesn't exist anymore with https://github.com/interledger/web-monetization-extension/pull/1085.

Storage is about more permanent things. `minRateOfPay` changes depending on website (tab) and what wallets they're using, and what wallets user can send money to.

Issue #1121, opened 2025-07-09
[BUG] Connect: amount input resets with default value


sidvishnoi:
### Steps to reproduce

1. Open connect wallet screen
2. Enter a wallet address
3. Quickly, update amount before wallet address is fetched

### Expected result

User's entered amount value is preserved

### Actual result

Amount is reset to the default value returned based on wallet address (https://github.com/interledger/web-monetization-extension/pull/1073).

### Additional context

Was trying to fix E2E tests, but they reset as amount is reset.

### Operating system

Linux

### Browsers

Chrome

### Extension version

v0.9.0.0 to nightly

Issue #957, opened 2025-03-04
Simplify ConnectWalletForm


sidvishnoi:
The form has become too complicated for a simple form with two fields. As mentioned in https://github.com/interledger/web-monetization-extension/pull/940.
Architect it a bit better, so new things are easier to add. Also, reduce re-renders happening too often (we've more props than state). React upgrade to v19 might help with  simplification too.

PR #1122, opened 2025-07-09
fix(ConnectWalletForm): avoid overriding user set amount

sidvishnoi:
Also prevent some re-renders, simplify form a bit

## Context

Closes https://github.com/interledger/web-monetization-extension/issues/1121
Part of https://github.com/interledger/web-monetization-extension/issues/957
Fixes e2e tests broken by https://github.com/interledger/web-monetization-extension/pull/1073

- Don't override amount if already set by user
   - E2E tests `fillPopup` couldn't set amounts different than `5.0` due to this
   - Make use of a `data-modified` attribute to detect if user modified input
- Refactor `ConnectWalletForm` a bit (extract `AmountInput`) and reduce some re-renders

### DIFF
--- a/src/pages/popup/components/ConnectWalletForm.tsx
+++ b/src/pages/popup/components/ConnectWalletForm.tsx
@@ -10,7 +10,12 @@ import {
   InputAmount,
   validateAmount,
 } from '@/pages/shared/components/InputAmount';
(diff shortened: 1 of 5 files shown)


sidvishnoi (commented):
test-e2e

PR #1125, opened 2025-07-10
fix(bg/wallet): improve new tab opening, avoid empty tabs

sidvishnoi:
## Context

We were opening empty tabs early, and then updating them with URL. On fast devices, this was alright, but on slower devices, user will see an empty tab open, then see it update.
Also, it made watching for page opens harder in E2E tests (we added that custom `waitForPage` helper).

- Make use of `onTabOpen` callback in functions that need to open a tab.
   - Instead of preemptively opening tab for use in functions, create one when needed, and let call-site know of it.
- Update E2E tests to remove the custom `waitForPage` helper.
- Also, instead of reusing tabs which failed on post-install connect screen in Safari, close extension tabs (include app page and webmonetization.org/welcome pages) to keep tab state clean. Open new tab after closing these extension-spefici tabs to reduce noise.
- Removed `reuseOrCreateTab` and replace it with `createTab`, `createTabIfNotExists` (not exists == if a tabId is undefined, or if tab with tabId is already closed) and `closeTabsByFilter` helpers.

### DIFF
--- a/src/pages/app/pages/PostInstall.tsx
+++ b/src/pages/app/pages/PostInstall.tsx
@@ -4,11 +4,8 @@ import {
   CaretDownIcon,
   ExternalIcon,
 } from '@/pages/shared/components/Icons';
-import {
-  getBrowserName,
-  getConnectWalletInfo,
-  type BrowserName,
-} from '@/shared/helpers';
+import { getBrowserName, type BrowserName } from '@/shared/helpers';
+import { getResponseOrThrow } from '@/shared/messages';
 import { useBrowser, useTranslation } from '@/app/lib/context';
 import { ConnectWalletForm } from '@/popup/components/ConnectWalletForm';
 import { cn } from '@/pages/shared/lib/utils';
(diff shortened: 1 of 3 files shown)


sidvishnoi (commented):
test-e2e
More tests should pass than last PR.

src/pages/popup/components/ConnectWalletForm.tsx
```diff
@@ -316,6 +316,7 @@ export const ConnectWalletForm = ({
         onAccept={async () => {
           autoKeyAddConsent.current = true;
           setShowConsent(false);
+          await sleep(200);
```
DarianM, line 319: why do we need this await ?

sidvishnoi: ConnectWalletForm is messy. We're waiting for the form inputs to get rendered once we hide the consent screen, so we can read input amount. It's hacky, until we fix ConnectWalletForm for good (https://github.com/interledger/web-monetization-extension/issues/957).

DarianM (approved):

PR #1127, opened 2025-07-11
fix(ConnectWalletForm): ensure walletInfo is there after consent

sidvishnoi:
## Context

After key-add consent, there's sometimes a re-render of ConnectWalletForm, which makes it lose the fetched wallet info, which is later using, leading to `cannot access property of null` errors. More pronounced in E2E tests.

Ensure `walletInfo` is available before use.


DarianM (commented):
test-e2e

DarianM (approved):

PR #1133, opened 2025-07-14
fix(bg): correctly resume sessions when revoked key is re-added

sidvishnoi:
## Context

When the key was revoked, and we came back to previous tab, the payment sessions didn't have the `minSendAmount`, so they weren't "usable", and we had "no sessions!!" error.


sidvishnoi (commented):
test-e2e
The `reconnectAutoKeyTestWallet.spec` should pass now.

DarianM (approved):

PR #1211, opened 2025-10-20
fix: remove fixed popup height to prevent scrollbar at high zoom

### LINKED ISSUE #1014: Popup has scrollbar with browser zoom is more than 100%
In your browser settings, set zoom level to a value larger than 100%. The popup will end up having a scrollbar, which isn't ideal, specially when zoom level isn't too large.

Screenshot from Edge with popup showing scrollbar at 110% zoom

ARYPROGRAMMER:
## Context

Popup has scrollbar when browser zoom >100%.

The popup uses a fixed height constraint that doesn't accommodate content scaling at zoom levels above 100%, causing unwanted scrollbars. This affects usability across different browsers and devices.

Closes #1014 

Remove the fixed h-popup height class from the MainLayout component to enable dynamic popup sizing based on content height. This ensures the popup window resizes automatically to fit content at all zoom levels without scrollbars.

Modified MainLayout.tsx to remove h-popup class from the root div

### DIFF
--- a/src/pages/popup/components/layout/MainLayout.tsx
+++ b/src/pages/popup/components/layout/MainLayout.tsx
@@ -2,16 +2,19 @@ import React from 'react';
 import { Header } from './Header';
 
 const Divider = () => {
-  return <div className="w-100 h-1 bg-divider-gradient" />;
+  return <div className="w-full h-px bg-divider-gradient" />;
 };
 
 export const MainLayout = ({ children }: { children: React.ReactNode }) => {
   return (
     // biome-ignore lint/correctness/useUniqueElementIds: main is unique, as at top-level layout
-    <div className="flex w-popup flex-col gap-4 px-6 py-4" id="main">
+    <div
+      id="main"
+      className="flex w-popup min-h-[480px] min-w-[360px] max-h-[90vh] h-auto flex-col gap-4 px-6 py-4 overflow-hidden"
+    >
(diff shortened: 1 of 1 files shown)


ARYPROGRAMMER:
> The problem with removing the height like this means the extension popup will keep changing size based on content, which isn't a user good experience.
> 
>  Screen.Recording.2025-10-21.at.14.12.38.mov

the other fix would involve some or the other hardcoded things, even though the connect button is dynamically positioned it would be the most optimal change [maybe add the button to the top]. in any case i am doing another change this is also good and solves the problem.

@sidvishnoi did the changes

sidvishnoi:
@ARYPROGRAMMER I see in the wallet info screen, there's a scrollbar even without browser zoom when the advanced section is open, and the overall UI looks squished.

image

Also, note info/warning pages have content moved to top instead of center:
image


I'm not sure of the way forward here. I have asked the team for some design direction. 

Meanwhile, would you like to work on some other issues? 1187 and 1188 are good options - I can assign either to you if you comment on those issues.

ARYPROGRAMMER:
yes pls you can assign me both @sidvishnoi ,the problem with this is that the design might be too complicated or might need re refractoring of the mainlayout for that per say. I still think h-popup removal was the easiest and comfortable thing to do but yes surely let's come back to this later

PR #1214, opened 2025-10-23
chore(deps): remove tsx dependency

### LINKED ISSUE #1187: Remove `tsx` dependency
Node.js v24 includes native support for running TS files with type-stripping. It also supports the `--watch` flag to re-run script on changes. We can remove `tsx` as dependency now and run our dev/build scripts directly with `node`.

We might need to enable in `tsconfig.json` to enable using only erasable types.

ARYPROGRAMMER:
## Context

Node.js v24 now supports native TypeScript execution with type-stripping and the `--watch` flag, eliminating the need for the `tsx` dependency.

This change removes `tsx` as a dev dependency and updates the build scripts to use `node` directly.

Closes #1187 

- Updated `package.json` scripts to use `node` instead of `tsx` for build and dev commands
- Removed `tsx` from devDependencies in `package.json`
- Updated documentation in `docs/DEVELOP.md` and `src/shared/logger.ts` to reference `node` instead of `tsx`

### DIFF
--- a/esbuild/dev.ts
+++ b/esbuild/dev.ts
@@ -1,7 +1,7 @@
 import { readFile } from 'node:fs/promises';
 import type { BuildOptions, Plugin as ESBuildPlugin } from 'esbuild';
-import { SERVE_PORTS, type BuildArgs, type Target } from './config';
-import { getPlugins } from './plugins';
+import { SERVE_PORTS, type BuildArgs, type Target } from './config.ts';
+import { getPlugins } from './plugins.ts';
 
 export const getDevOptions = ({
   outDir,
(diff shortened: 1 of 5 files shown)


ARYPROGRAMMER:
for review: @sidvishnoi

package.json
```diff
@@ -82,7 +82,6 @@
     "sade": "^1.8.1",
     "tailwindcss": "^3.4.18",
     "tinyspy": "^4.0.4",
-    "tsx": "^4.20.6",
```
sidvishnoi: @ARYPROGRAMMER can you run `pnpm i` so the `pnpm-lock.yaml` gets updated as well? Include that in the PR.

sidvishnoi (changes requested):
@ARYPROGRAMMER Please bring on those `.ts` import changes to this PR, to ensure dev/build script works.

PR #1358, opened 2026-04-15
fix(bg): success redirect only after reconnect post-process done

sidvishnoi:
## Context

Follow-up for https://github.com/interledger/web-monetization-extension/pull/1357 (UX improvement)

Redirect user to success only after reconnect post-processing is finished.

### DIFF
--- a/src/background/services/background.ts
+++ b/src/background/services/background.ts
@@ -264,9 +264,16 @@ export class Background {
         }
 
         case 'RECONNECT_WALLET': {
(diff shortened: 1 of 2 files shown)


DarianM (approved):

sidvishnoi:
Thinking about this more in context of https://github.com/interledger/web-monetization-extension/pull/1329. I think the delay can be substantial for user to be waiting for success-redirect. Let's not do this.

PR #1362, opened 2026-04-22
perf: replace crypto-browserify with custom lite impl

### LINKED ISSUE #1206: Look into getting rid of `crypto-browserify`
[`crypto-browserify`](https://github.com/browserify/crypto-browserify) has enough methods that are available in SubtleCrypto API, readily available in the browser. Check what specific functions we need this particular library for, and see if we can have a thin wrapper over SubtleCrypto to get the same polyfill effect.

Presently, that this library a lot of outdated packages to our bundle, and I see a lot of dependabot security alerts due to this.

Relates to https://github.com/interledger/open-payments-node/issues/4.

sidvishnoi:
## Context

Closes #1206.

```diff
-   dist/chrome/background/background.js                     823.9kb
+   dist/chrome/background/background.js                     526.1kb
```

E2E confirm working for Chromium, Edge.
Tested on latest Firefox & Safari manually.

### DIFF
--- a/scripts/esbuild/polyfill/lite-crypto.ts
+++ b/scripts/esbuild/polyfill/lite-crypto.ts
@@ -8,14 +8,10 @@ class Hash {
 
   constructor(algorithm: string) {
     const algo = algorithm.toLowerCase().replace(/-/g, '');
-    switch (algo) {
-      case 'sha512':
-        this.instance = sha512.create();
-        break;
-      default:
-        throw new Error(
-          `Algorithm ${algorithm} is not supported in sync mode.`,
-        );
+    if (algo === 'sha512') {
+      this.instance = sha512.create();
+    } else {
+      throw new Error(`lite-crypto: ${algorithm} is not implemented.`);
     }
   }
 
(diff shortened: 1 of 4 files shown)


scripts/esbuild/polyfill/lite-crypto.ts
```diff
@@ -0,0 +1,62 @@
...
+      return btoa(String.fromCharCode(...hash));
+    }
+
+    // Default: Return Uint8Array (Node returns a Buffer, which is a Uint8Array subclass)
+    return hash;
+  }
+}
+
+export const createHash = (algo: string) => {
+  return new Hash(algo);
```
DarianM, line 57: if the build conditions ever shift or a new dependency that pulls in like `createHash('sha256')`, this polyfill will throw

DarianM (approved):

scripts/esbuild/polyfill/lite-crypto.ts
sidvishnoi: Yes. Let me update error message a bit. Our tests will fail then of course.

PR #1388, opened 2026-06-22
fix(deps): don't override `structured-headers` version

sidvishnoi:
## Context

The E2E failures in https://github.com/interledger/web-monetization-extension/pull/1387 weren't transient (Nightly has been failing since the updates, some transient issues hid the real breakge). `http-message-signatures` bumped their `structured-headers` version to v2, while we were forcing v1, so OP integration broke.

Don't restrict `structured-headers` dependency version, let the packages define their versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,8 +94,7 @@
       "safe-buffer": "5.2.1",
       "string_decoder": "1.3.0",
       "readable-stream": "2.3.8",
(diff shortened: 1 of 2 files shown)


sidvishnoi:
E2E failures now are due to transient/network conditions.

DarianM (approved):

PR #1409, opened 2026-07-06
feat(popup/RateOfPay): configure current site's rate

### LINKED ISSUE #1382: UI: Implement adding a per-site rate
As a user, I want to assign a different pay rate than the default to my favourite websites so that I can support them more than sites I visit casually.

Design: Available on Figma
Acceptance Criteria: 

- Users can assign a higher rate of pay to a specific site
- Users can assign a lower rate of pay to a specific site 
- Users can see the rate applied to the page they're on

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/1382
Part of https://github.com/interledger/web-monetization-extension/issues/1345
Part of https://github.com/interledger/web-monetization-extension/issues/1297 (story)

- In SettingsRateOfPay, add form/inputs to set site-specific rate (for current tab).
- Add "Add exception" (when current tab/site doesn't have an exception rate set) and "manage exceptions" (when a rate for current tab/site is set)
   - Manage exceptions button takes user to "Settings -> other", to be developed separately.
- Fix one place in background that didn't add tab's rate of pay but override rest of the data (this has some performance optimization opportunity; for later)
- On the home page, added a "*" symbol to denote a custom rate is active; will be re-designed in upcoming PRs.

https://github.com/user-attachments/assets/778a5593-a8b9-44f5-99d0-924eef06d350

### DIFF
--- a/src/pages/popup/components/Settings/RateOfPay.tsx
+++ b/src/pages/popup/components/Settings/RateOfPay.tsx
@@ -123,7 +123,7 @@ export const RateOfPayComponent = ({
       >
         <RateOfPayInput
           id="rateOfPay"
-          label="Default rate"
+          label={t('settings_rate_label_inputDefaultRate')}
           onRateChange={onRateChange}
           rateOfPay={rateOfPay}
           maxRateOfPay={maxRateOfPay}
@@ -187,6 +187,7 @@ const SiteRateOfPayInput = ({
 }: {
   onRateChange: Props['onSiteRateChange'];
 }) => {
+  const t = useTranslation();
   const { rateOfPay, maxRateOfPay, walletAddress, tab } = usePopupState();
 
   const hostname = new URL(tab.url).hostname;
@@ -206,7 +207,9 @@ const SiteRateOfPayInput = ({
           type="button"
           className="p-1 text-alt"
         >
-          <span className="sr-only">Remove exception</span>
+          <span className="sr-only">
+            {t('settings_rate_action_removeException')}
+          </span>
           <IconTrash className="size-4" />
         </button>
       </div>
--- a/src/pages/popup/components/Settings/RateOfPay.tsx
+++ b/src/pages/popup/components/Settings/RateOfPay.tsx
@@ -4,7 +4,7 @@ import { SwitchButton } from '@/pages/shared/components/ui/Switch';
 import { Button } from '@/pages/shared/components/ui/Button';
 import { InputAmountMemoized as InputAmount } from '@/pages/shared/components/InputAmount';
 import { IconTrash } from '@/pages/shared/components/Icons';
-import { debounceAsync } from '@/shared/helpers';
+import { debounceAsync, normalizeHostname } from '@/shared/helpers';
 import { cn, formatNumber, roundWithPrecision } from '@/pages/shared/lib/utils';
 import { useMessage, useTelemetry, useTranslation } from '@/popup/lib/context';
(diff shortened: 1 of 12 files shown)


src/pages/popup/components/Settings/RateOfPay.tsx
```diff
@@ -80,27 +108,124 @@ export const RateOfPayComponent = ({ onRateChange, toggle }: Props) => {
...
+      <div
+        className={cn(
+          'border p-3 rounded-lg',
+          continuousPaymentsEnabled ? 'border-popup' : 'border-base',
+          continuousPaymentsEnabled && !tab.rateOfPay ? 'bg-popup-light' : '',
         )}
+      >
+        <RateOfPayInput
+          id="rateOfPay"
+          label="Default rate"
```
MichaelKohler: Let's move this to translations as well.

src/pages/popup/components/Settings/RateOfPay.tsx
```diff
@@ -80,27 +108,124 @@ export const RateOfPayComponent = ({ onRateChange, toggle }: Props) => {
...
 
+const SiteRateOfPayInput = ({
+  onRateChange,
+}: {
+  onRateChange: Props['onSiteRateChange'];
+}) => {
+  const { rateOfPay, maxRateOfPay, walletAddress, tab } = usePopupState();
+
+  const hostname = new URL(tab.url).hostname;
+  const site = hostname.replace(/^www\./, '');
```
MichaelKohler: It's possible to have different exceptions for `example.com` and `www.example.com`, right? Might this be confusing if we remove the `www` then? I acknowledge that only either one will be shown as it is, but that might change with the "Manage exceptions" parts, right?

src/pages/popup/components/Settings/RateOfPay.tsx
```diff
@@ -80,27 +108,124 @@ export const RateOfPayComponent = ({ onRateChange, toggle }: Props) => {
...
+          className="flex items-center px-2 font-medium leading-6 text-medium"
+        >
+          {site} (exception)
+        </label>
+        <button
+          onClick={() => onRateChange({ rate: null, hostname })}
+          type="button"
+          className="p-1 text-alt"
+        >
+          <span className="sr-only">Remove exception</span>
```
MichaelKohler: Let's also move this to translations

src/pages/popup/components/Settings/RateOfPay.tsx
MichaelKohler: This also goes for other usages of `normalizeHostname`. And we could also use `normalizeHostname` here if that's desired.

MichaelKohler (approved):
Nice work! Made a few comments though none of these are blocking.

I noticed that when trying to increase the rate above the maximum with the input arrows, there is no indication why you can't go above the maximum. However when entering something higher than the maximum, there is an error message:

Image

src/pages/popup/components/Settings/RateOfPay.tsx
sidvishnoi: Yes, my understanding from discussions is www will be a special case - that we'll always normalize it. I'll update so other places also use `normalizeHostname`.

sidvishnoi:
> I noticed that when trying to increase the rate above the maximum with the input arrows, there is no indication why you can't go above the maximum. However when entering something higher than the maximum, there is an error message:

Let's address this separately - as the component gets used in lots of places.

Issue #1381, opened 2026-05-14
UI: Implement a list of site-specific rates


RabebOthmani:
As a user, I want to view and manage all my site-specific rates in one place so that I have a clear picture of my monetization preferences.

Design: available on Figma 
Acceptance criteria: 

- The list of site-specific rates is located under settings 
- Users have the ability to add a new entry to the list 
- Users have the ability to change the rate from within the list
- Users have the ability to delete an entry from the list

PR #1411, opened 2026-07-08
refactor(popup/Settings/Other): extract into components

sidvishnoi:
## Context

Part of https://github.com/interledger/web-monetization-extension/issues/1381

- Move `DataCollectionSettings` out of `Settings`
- Create a `SettingsAccordion` component for multiple settings (will be reused with a future RateOfPayManage component)

(E2E test failure is unrelated; something up with Interledger wallet)

### DIFF
--- a/src/pages/popup/components/Settings/Settings.tsx
+++ b/src/pages/popup/components/Settings/Settings.tsx
@@ -1,103 +1,41 @@
 import React from 'react';
-import { SwitchButton } from '@/pages/shared/components/ui/Switch';
 import { CaretDownIcon } from '@/pages/shared/components/Icons';
-import {
-  useBrowser,
-  useBrowserInfo,
-  useTelemetry,
-  useTranslation,
-} from '@/pages/shared/lib/context';
-import { dispatch, usePopupState } from '@/popup/lib/store';
-import { useMessage } from '@/popup/lib/context';
-import type { Browser } from '@/shared/browser';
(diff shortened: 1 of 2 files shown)


src/pages/popup/components/Settings/Settings.tsx
```diff
@@ -1,103 +1,41 @@
...
-import {
-  useBrowser,
-  useBrowserInfo,
-  useTelemetry,
-  useTranslation,
-} from '@/pages/shared/lib/context';
-import { dispatch, usePopupState } from '@/popup/lib/store';
-import { useMessage } from '@/popup/lib/context';
-import type { Browser } from '@/shared/browser';
+import { useTranslation } from '@/app/lib/context';
```
DarianM: nit. i noticed we use `@/app/lib/context` or `@/popup/lib/context` for useTranslation and both compile fine.
but this is the only `/app/*` import i see in this file and  it makes me think the popup depend on the app page's context module with this import

DarianM (approved):

src/pages/popup/components/Settings/Settings.tsx
sidvishnoi: Oops, editor auto import.

sidvishnoi:
```suggestion
import { useTranslation } from '@/popup/lib/context';
```

PR #1417, opened 2026-07-10
feat(bg/outgoingPaymentGrant): get spent grant amounts

### LINKED ISSUE #737: [BUG] extension balance inconsistency & unable to pay one-time
### Steps to reproduce

1. Navigated to a monetized web page (https://sidvishnoi.com/test/wm/).
2. Conditions: testing with a `ilp.interledger-test` USD payment pointer connected to the extension, with continuous payments enabled, the hourly rate set to $1, and monthly renewal enabled.
3. Started with the extension showing a $0.41 available balance.
4. Tried a few unsuccessful one-time payments for $0.38, $0.34 & $0.30. Each attempt failed, with the extension showing the message `Insufficient funds to complete payment.`
6. Navigated to Settings > Budget to check the available balance, now at $0.39
7. Returned to the landing page and paid $0.20, which was successful **but** the extension's remaining balance didn't take the successful one-time payment into account (now showing as $0.37).

### Expected result

After the extension shows a message that a payment succeeded, I expected for the extension's balance to decrease (more-or-less immediately). 

### Actual result

There was a fairly long delay (a few minutes, and few actions later) before the extension showed the lower balance.

### Screenshots or videos

https://www.loom.com/share/e03c96b43ad24143936f8aa1713c1689?sid=b2616454-6383-420e-bed1-2ad03052b448

### Additional context

_No response_

### Operating system

Windows

### Operating system version

10

### Browsers

Chrome

### Browser version

Vivaldi 7.0.3495.18 (Stable channel) (64-bit) 

### Extension version

Nightly 2024.11.26

MichaelKohler:
## Context

To more reliably update the balance, we want to introduce usage of the Open Payments [/outgoing-payment-grant](https://openpayments.dev/apis/resource-server/operations/get-outgoing-payment-grant/) endpoint. 

Part of #737.

This PR introduces the function to fetch the spent amounts, as well as a new storage property that indicates whether the current wallet address' resource server does support this new endpoint. This will then in a follow-up be used to guard polling the balance on the popup as well as not polling outgoing payments.

Follow-up PR will be opened once I have looked closer into the popup lifecycle, therefore already this PR, keeping it small.

## Testing

* Tested request towards Testnet
* Tested successful saving in storage
* Tested against GateHub (not supported yet), also correctly saved as "false"
* Tested reset on wallet disconnect

### DIFF
--- a/src/background/services/outgoingPaymentGrant.ts
+++ b/src/background/services/outgoingPaymentGrant.ts
@@ -14,7 +14,10 @@ import {
 } from '@interledger/open-payments';
 import type { Browser, Tabs } from 'webextension-polyfill';
 import type { Cradle } from '@/background/container';
-import { ACCEPT_GRANT_TIMEOUT, MAX_GET_GRANT_SPENT_AMOUNTS_RETRIES } from '@/background/config';
+import {
+  ACCEPT_GRANT_TIMEOUT,
+  MAX_GET_GRANT_SPENT_AMOUNTS_RETRIES,
+} from '@/background/config';
 import { OPEN_PAYMENTS_REDIRECT_URL } from '@/shared/defines';
 import { ensureEnd, ErrorWithKey, withResolvers } from '@/shared/helpers';
 import {
--- a/src/background/services/outgoingPaymentGrant.ts
+++ b/src/background/services/outgoingPaymentGrant.ts
@@ -260,6 +260,27 @@ export class OutgoingPaymentGrantService {
     }
   }
 
+  async getGrantSpentAmounts(walletAddress: WalletAddress) {
+    try {
+      const spentAmounts = await this.openPaymentsService.client.outgoingPayment.getGrantSpentAmounts({
+        url: walletAddress.resourceServer,
+        accessToken: this.accessToken,
+      });
+
+      await this.storage.set({
+        supportsGrantSpentAmounts: true,
+      });
+
+      return spentAmounts;
+    } catch (error) {
+      this.logger.debug('Resource server does not support grant spent amounts endpoint', error);
+
+      await this.storage.set({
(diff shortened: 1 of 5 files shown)


src/background/services/outgoingPaymentGrant.ts
```diff
@@ -260,6 +260,27 @@ export class OutgoingPaymentGrantService {
...
+  async getGrantSpentAmounts(walletAddress: WalletAddress) {
+    try {
+      const spentAmounts = await this.openPaymentsService.client.outgoingPayment.getGrantSpentAmounts({
+        url: walletAddress.resourceServer,
+        accessToken: this.accessToken,
+      });
+
+      await this.storage.set({
+        supportsGrantSpentAmounts: true,
+      });
```
sidvishnoi, line 290: I think we should store it in memory, so when the wallet starts supporting this endpoint (like the next time browser opens), we can start using it. Adding this state to storage feels more permanent.

MichaelKohler: As discussed, I will introduce a weekly check instead.

src/background/services/outgoingPaymentGrant.ts
```diff
@@ -260,6 +265,50 @@ export class OutgoingPaymentGrantService {
...
+
+      return spentAmounts;
+    } catch (error) {
+      if (isTokenExpiredError(error)) {
+        await this.rotateToken();
+        return await this.getGrantSpentAmounts(walletAddress, retry + 1);
```
MichaelKohler, line 296: The retry mechanism here feels quite defensive, but if something goes wrong and it does not throw, we'd end up in a very bad state.

sidvishnoi (approved):
:shipit: First commit 🚀

PR #1418, opened 2026-07-10
perf(ConnectWalletForm): prevent unnecessary rerenders

sidvishnoi:
## Context

Part of https://github.com/interledger/web-monetization-extension/issues/957

`ConnectWalletForm` re-rendered its entire tree every few seconds while a wallet connection was in progress, because the connect status (which updates frequenty during connect process) was passed as a prop. We can't simply use `usePopupState` in the form, as the form is used in app as well (where we need `useAppState`).

The problem mostly shows with how we've to keep so many checks with wallet address input, as re-renders keep triggering fetch/validate wallet address calls.

- Introduced a `ConnectStateContext` that carries the transient connect state, consumed only by two small leaf components (`ConnectStateErrorSync` for syncing failure/cancel errors into local error state, and `ConnectSubmitLoadingText` for the submit button's progress text).. instead of the whole form.
- Wrapped `ConnectWalletForm` in `React.memo` so it actually skips re-rendering when its (now-stable) props don't change. The context split alone wasn't enough without this, since the parent page still re-render on every tick (when transientState changed in useAppState/usePopupState)
- Exported some shared helpers used in both App and Popup (`useConnectWalletFormActions`, `saveValue` etc.)

Follow up: `ConnectWalletForm` is 750 lines long. Need to simplify and make it smaller - extract components, functions; and make it more testable. Also look for further flow simplifications.

### DIFF
--- a/src/pages/app/pages/PostInstall.tsx
+++ b/src/pages/app/pages/PostInstall.tsx
@@ -6,9 +6,14 @@ import {
   ExternalIcon,
 } from '@/pages/shared/components/Icons';
 import { isConsentRequired, type BrowserName } from '@/shared/helpers';
-import { getResponseOrThrow } from '@/shared/messages';
 import { useBrowser, useBrowserInfo, useTranslation } from '@/app/lib/context';
-import { ConnectWalletForm } from '@/popup/components/ConnectWalletForm';
+import {
+  ConnectStateContext,
+  ConnectWalletForm,
+  getSavedValues,
+  saveValue,
+  useConnectWalletFormActions,
+} from '@/popup/components/ConnectWalletForm';
 import { cn } from '@/pages/shared/lib/utils';
 import { useMessage } from '@/app/lib/context';
 import { useAppState } from '@/app/lib/store';
@@ -455,13 +460,8 @@ function StepConnectWallet({
 }: {
   selectedWallet: WalletOption;
 }) {
-  const message = useMessage();
   const t = useTranslation();
-  const {
-    transientState: { connect: connectState },
-    connected,
-    publicKey,
-  } = useAppState();
+  const { connected } = useAppState();
 
   if (connected) {
     return (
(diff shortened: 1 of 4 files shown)


src/pages/app/pages/PostInstall.tsx
```diff
@@ -476,33 +476,45 @@ function StepConnectWallet({
...
-        getWalletInfo={(walletAddressUrl) =>
-          message
-            .send('GET_CONNECT_WALLET_ADDRESS_INFO', walletAddressUrl)
-            .then(getResponseOrThrow)
-        }
-        walletAddressPlaceholder={selectedWallet.walletAddressPlaceholder}
-        connectWallet={(data) => message.send('CONNECT_WALLET', data)}
-        clearConnectState={() => message.send('RESET_CONNECT_STATE')}
-      />
+      {/* @ts-expect-error we know, it's complicated */}
```
MichaelKohler, line 504: Out of scope: This is indeed complicated.

```
diff --git i/src/shared/types.ts w/src/shared/types.ts
index c21f2f5..bd359ed 100644
--- i/src/shared/types.ts
+++ w/src/shared/types.ts
@@ -250,9 +250,9 @@ export type DeepNonNullable<T> = {
   [P in keyof T]?: NonNullable<T[P]>;
 };
 
-export type DeepReadonly<T> = {
-  readonly [P in keyof T]: DeepReadonly<T[P]>;
-};
+export type DeepReadonly<T> = T extends object
+  ? { readonly [P in keyof T]: DeepReadonly<T[P]> }
+  : T;
 
 export type DeepPartial<T> = T extends object
   ? { [P in keyof T]?: DeepPartial<T[P]> }

```

Would help though.

MichaelKohler (approved):

src/pages/app/pages/PostInstall.tsx
sidvishnoi: I was thinking of creating a small type-assertion helper for this, or maybe modify the expected type for state. Let me address this as a follow-up.

PR #1419, opened 2026-07-10
refactor(ConnectWalletForm): simplify, extract into independent components

### LINKED ISSUE #957: Simplify ConnectWalletForm

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/957

See commit messages. Review by commit perhaps, instead of overall.
Makes the components more self-contained, fixes some state related issues, makes overall flow simpler to follow (_it's just a form_!).
Didn't split into files, as makes diff much larger (after merge) and somehow crashing Biome linter as well.

Only behavior change: remove autofocus on amount input on paste in wallet address.

### DIFF
--- a/src/pages/popup/components/ConnectWalletForm.tsx
+++ b/src/pages/popup/components/ConnectWalletForm.tsx
@@ -475,7 +475,8 @@ function WalletAddressInput({
         onWalletInfoChange(walletInfo);
         return walletInfo;
       } catch (err) {
-        setError(toErrorInfo(err.message));
+        const msg = err instanceof Error ? err.message : String(err);
+        setError(toErrorInfo(msg));
         return null;
       } finally {
         setIsValidating(false);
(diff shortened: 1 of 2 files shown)


src/pages/popup/components/ConnectWalletForm.tsx
sidvishnoi, line 1: Biome crash still triggering. Will figure out later.

sidvishnoi: Triggers only when files are split - so something we Biome's module resolver.

sidvishnoi: So, reverting the file split from this PR.

MichaelKohler (approved):

PR #1423, opened 2026-07-13
feat(bg/grantBalance): update remaining balance regularly and on popup open

### LINKED ISSUE #737: [BUG] extension balance inconsistency & unable to pay one-time

MichaelKohler:
## Context

Part of #737.

With this change we now register the regular balance update when opening the popup. Based on the continuous payment setting the interval period is changed. We also regularly re-check for grant spent amount support and do not poll if it is supported.

### DIFF
--- a/src/background/services/grantBalance.ts
+++ b/src/background/services/grantBalance.ts
@@ -1,0 +1,183 @@
+import type {
+  WalletAddress,
+  OutgoingPaymentGrantSpentAmounts,
+} from '@interledger/open-payments';
+import type { Cradle } from '@/background/container';
+import {
+  GRANT_SPENT_AMOUNTS_SUPPORT_RECHECK_ALARM,
+  GRANT_SPENT_AMOUNTS_SUPPORT_RECHECK_INTERVAL_MS,
+} from '@/background/config';
+import { Timeout } from '@/shared/helpers';
+import { isNotFoundError } from '@/background/services/openPayments';
+import { onPopupOpen } from '@/background/utils';
+
+export class GrantBalanceService {
+  private storage: Cradle['storage'];
+  private logger: Cradle['logger'];
+  private outgoingPaymentGrantService: Cradle['outgoingPaymentGrantService'];
+  private browser: Cradle['browser'];
+  private balanceUpdateTimeout: Timeout | null = null;
+
+  constructor({
+    storage,
+    logger,
+    outgoingPaymentGrantService,
+    browser,
+  }: Cradle) {
+    this.storage = storage;
+    this.logger = logger;
+    this.outgoingPaymentGrantService = outgoingPaymentGrantService;
+    this.browser = browser;
+
+    this.initialize();
+  }
+
+  private initialize() {
+    void this.registerBalanceUpdateHandler();
+  }
+
+  start() {
+    this.browser.alarms.onAlarm.addListener(async (alarm) => {
+      if (alarm.name !== GRANT_SPENT_AMOUNTS_SUPPORT_RECHECK_ALARM) return;
+
+      const supported = await this.checkGrantSpentAmountsSupport();
+      if (supported) {
+        this.browser.alarms.clear(GRANT_SPENT_AMOUNTS_SUPPORT_RECHECK_ALARM);
+      }
+    });
+
+    void this.checkGrantSpentAmountsSupport();
+  }
+
+  private async getGrantSpentAmounts(
+    walletAddress: WalletAddress,
+  ): Promise<OutgoingPaymentGrantSpentAmounts | undefined> {
+    try {
+      const spentAmounts =
+        await this.outgoingPaymentGrantService.getGrantSpentAmounts(
+          walletAddress,
+        );
+
+      await this.storage.set({
+        supportsGrantSpentAmounts: {
+          supported: true,
+          lastCheckedAt: Date.now(),
+        },
+      });
+
+      return spentAmounts;
+    } catch (error) {
+      if (isNotFoundError(error)) {
+        this.logger.debug(
+          'Resource server does not support grant spent amounts endpoint',
+          error,
+        );
+
+        await this.storage.set({
+          supportsGrantSpentAmounts: {
+            supported: true,
+            lastCheckedAt: Date.now(),
+          },
+        });
+      } else {
+        const { supportsGrantSpentAmounts } = await this.storage.get([
+          'supportsGrantSpentAmounts',
+        ]);
+
+        await this.storage.set({
+          supportsGrantSpentAmounts: {
+            supported: supportsGrantSpentAmounts?.supported,
+            lastCheckedAt: Date.now(),
+          },
+        });
+
+        throw error;
+      }
+    }
+  }
+
+  private async checkGrantSpentAmountsSupport() {
+    if (!this.outgoingPaymentGrantService.isAnyGrantUsable) return;
+
+    const { supportsGrantSpentAmounts, walletAddress } = await this.storage.get(
+      ['supportsGrantSpentAmounts', 'walletAddress'],
+    );
+
+    if (supportsGrantSpentAmounts?.supported || !walletAddress) return;
+
+    const existingGrantSpentAmountAlarm = await this.browser.alarms.get(
+      GRANT_SPENT_AMOUNTS_SUPPORT_RECHECK_ALARM,
+    );
+
+    if (!existingGrantSpentAmountAlarm) {
+      this.browser.alarms.create(GRANT_SPENT_AMOUNTS_SUPPORT_RECHECK_ALARM, {
+        when:
+          (supportsGrantSpentAmounts?.lastCheckedAt ?? 0) +
+          GRANT_SPENT_AMOUNTS_SUPPORT_RECHECK_INTERVAL_MS,
+        periodInMinutes:
+          GRANT_SPENT_AMOUNTS_SUPPORT_RECHECK_INTERVAL_MS / 60 / 1000,
+      });
+    }
+
+    const spentAmounts = await this.getGrantSpentAmounts(walletAddress);
+    return spentAmounts?.spentDebitAmount !== undefined;
+  }
+
+  private async registerBalanceUpdateHandler() {
+    const { supportsGrantSpentAmounts } = await this.storage.get([
+      'supportsGrantSpentAmounts',
+    ]);
+
+    if (!supportsGrantSpentAmounts?.supported) {
+      return;
+    }
+
+    const updateBalance = async () => {
+      await this.saveUpdatedBalance();
+
+      const { continuousPaymentsEnabled } = await this.storage.get([
+        'continuousPaymentsEnabled',
+      ]);
+      const timeoutMs = continuousPaymentsEnabled
+        ? 1 * 60 * 1000
+        : 5 * 60 * 1000;
+
+      this.balanceUpdateTimeout = new Timeout(timeoutMs, async () => {
+        await this.saveUpdatedBalance();
+
+        if (this.balanceUpdateTimeout) {
+          this.balanceUpdateTimeout.reset(timeoutMs);
+        }
+      });
+    };
+
+    const unregisterTimeout = async () => {
+      if (this.balanceUpdateTimeout) {
+        this.balanceUpdateTimeout.clear();
+        this.balanceUpdateTimeout = null;
+      }
+    };
+
+    onPopupOpen(this.browser, updateBalance, unregisterTimeout);
+  }
+
+  private async saveUpdatedBalance() {
+    const { walletAddress } = await this.storage.get(['walletAddress']);
+
+    if (!walletAddress) return;
+
+    try {
+      const amounts = await this.getGrantSpentAmounts(walletAddress);
+
+      if (!amounts?.spentDebitAmount) return;
+
+      this.storage.setSpentAmount(
+        this.outgoingPaymentGrantService.grantType,
+        amounts.spentDebitAmount.value,
+      );
+    } catch (error) {
+      this.logger.warn('Background balance update failed', error);
+      this.balanceUpdateTimeout?.clear();
+    }
+  }
+}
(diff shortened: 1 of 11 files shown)


src/background/services/outgoingPaymentGrant.ts
```diff
@@ -309,6 +316,65 @@ export class OutgoingPaymentGrantService {
...
+  }
+
+  async registerBalanceUpdateHandler() {
+    const { continuousPaymentsEnabled } = await this.storage.get([
+      'continuousPaymentsEnabled',
+    ]);
+
+    const FIVE_MINUTES = 5 * 60 * 1000;
+    const ONE_MINUTE = 60 * 1000;
+    const timeoutMs = continuousPaymentsEnabled ? ONE_MINUTE : FIVE_MINUTES;
```
MichaelKohler: Does it make sense to use 1 and 5 here?

src/background/utils.ts
```diff
@@ -278,11 +281,20 @@ export const createTabIfNotExists = async (
...
+  closeCallback?: () => Promise<void>,
 ) => {
   const listener: OnConnectCallback = (port) => {
     if (port.name !== BACKGROUND_TO_POPUP_CONNECTION_NAME) return;
     if (port.error) return;
 
+    if (closeCallback) {
+      const disconnectListener: OnDisconnectCallback = () => {
+        void closeCallback();
+        port.onDisconnect.removeListener(disconnectListener);
```
sidvishnoi: Should this be called at line 303 as well?

src/background/services/outgoingPaymentGrant.ts
```diff
@@ -309,6 +316,65 @@ export class OutgoingPaymentGrantService {
...
+    } catch (error) {
+      this.logger.debug('Background balance update failed', error);
+
+      if (this.balanceUpdateTimeout) {
+        this.balanceUpdateTimeout.clear();
+      }
+    }
+  }
+
+  async registerBalanceUpdateHandler() {
```
sidvishnoi: Do you think it could be helpful to add this balance update (along with existing one) to a separate service?

src/background/services/outgoingPaymentGrant.ts
sidvishnoi: I feel we might want to update more often when popup is open, specially if an outgoing payment was made recently? I'm not sure, but 1min/5min sounds reasonable in general.

src/background/services/outgoingPaymentGrant.ts
```diff
@@ -309,6 +316,65 @@ export class OutgoingPaymentGrantService {
     }
   }
 
+  private async saveUpdatedBalance() {
+    const walletInfo = await this.storage.get(['walletAddress']);
+
+    if (!walletInfo.walletAddress) return;
```
sidvishnoi: Nit:
```suggestion
    const { walletAddress } = await this.storage.get(['walletAddress']);
    if (!walletAddress) return;
```

src/background/services/outgoingPaymentGrant.ts
```diff
@@ -309,6 +316,65 @@ export class OutgoingPaymentGrantService {
...
+      this.storage.updateSpentAmount(
+        this.grantType,
+        amounts.spentDebitAmount.value,
+      );
+    } catch (error) {
+      this.logger.debug('Background balance update failed', error);
+
+      if (this.balanceUpdateTimeout) {
+        this.balanceUpdateTimeout.clear();
+      }
```
sidvishnoi: Nit:
```suggestion
      this.balanceUpdateTimeout?.clear();
```

src/background/services/outgoingPaymentGrant.ts
```diff
@@ -309,6 +316,65 @@ export class OutgoingPaymentGrantService {
...
+        'Updating balance with grant spentDebitAmount',
+        amounts.spentDebitAmount.value,
+      );
+
+      this.storage.updateSpentAmount(
+        this.grantType,
+        amounts.spentDebitAmount.value,
+      );
+    } catch (error) {
+      this.logger.debug('Background balance update failed', error);
```
sidvishnoi:
```suggestion
      this.logger.warn('Background balance update failed', error);
```

src/background/services/outgoingPaymentGrant.ts
```diff
@@ -309,6 +316,65 @@ export class OutgoingPaymentGrantService {
...
+      const amounts = await this.getGrantSpentAmounts(walletInfo.walletAddress);
+
+      if (!amounts?.spentDebitAmount) return;
+
+      this.logger.debug(
+        'Updating balance with grant spentDebitAmount',
+        amounts.spentDebitAmount.value,
+      );
+
+      this.storage.updateSpentAmount(
```
sidvishnoi: This has worked well with existing "on payment create update balance", where it checks the "lowest" balance. But given `getGrantSpentAmounts` is source of truth, we might want to get rid of "enqueue" in this case.

The enqueue is there to prevent race conditions - as there can be multiple payments in parallel, and we only want the balance to be _reducing_ (not increasing/decreasing randomly) in the UI.

Can add an `immediate` flag to `updateSpentAmount` I guess.

sidvishnoi:
The E2E failures are genuine.

src/background/services/outgoingPaymentGrant.ts
```diff
@@ -309,6 +316,65 @@ export class OutgoingPaymentGrantService {
...
+  }
+
+  async registerBalanceUpdateHandler() {
+    const { continuousPaymentsEnabled } = await this.storage.get([
+      'continuousPaymentsEnabled',
+    ]);
+
+    const FIVE_MINUTES = 5 * 60 * 1000;
+    const ONE_MINUTE = 60 * 1000;
+    const timeoutMs = continuousPaymentsEnabled ? ONE_MINUTE : FIVE_MINUTES;
```
sidvishnoi: What happens when continuous payments are toggled after this registration is called? The timeout needs to be dynamic on that as well. For example, maybe the callback can set next `timeoutMs` instead of setting it once, depending on current `continuousPaymentsEnabled` state.

src/background/services/outgoingPaymentGrant.ts
MichaelKohler: The 2 new functions could very well fit in a new balance service, tough I'd probably avoid having a service with two functions in it. Guess this would depend on how much of the existing logic we'd port over. The existing balance update is just a few lines. We could consider moving the outgoing payment polling from `PaymentManager` as well, but that does not fit nicely and might be confusing to have this logic mixed with the balance.

sidvishnoi: Let's keep within grant service for now then.

src/background/services/paymentManager.ts
```diff
@@ -261,28 +261,36 @@ export class PaymentManager {
...
-          return outgoingPayments.get(sessionId);
-        }),
-    );
+    const { supportsGrantSpentAmounts } = await this.deps.storage.get([
+      'supportsGrantSpentAmounts',
+    ]);
+
+    const pollingResults = supportsGrantSpentAmounts
+      ? []
+      : await Promise.allSettled(
```
sidvishnoi: We do need this polling - not just for balance update. For example, if user doesn't have enough balance in account, this polling will show a warning/error message to user.

src/shared/types.ts
```diff
@@ -123,6 +123,9 @@ export interface Storage {
    */
   supportsGrantSpentAmounts?: boolean;
 
+  /** Timestamp of the last time we checked grant spent amounts support. */
+  supportsGrantSpentAmountsCheckedAt?: number;
```
sidvishnoi: Can we store these two in a single key as they're very related?

src/background/services/outgoingPaymentGrant.ts
```diff
@@ -302,13 +320,110 @@ export class OutgoingPaymentGrantService {
...
+      );
+    } catch (error) {
+      this.logger.warn('Background balance update failed', error);
+      this.balanceUpdateTimeout?.clear();
+    }
+  }
+
+  async registerBalanceUpdateHandler() {
+    const FIVE_MINUTES = 5 * 60 * 1000;
+    const ONE_MINUTE = 60 * 1000;
```
sidvishnoi: Maybe rename these something like `INTERVAL_POPUP_OPEN` and `INTERVAL_POPUP_CLOSED`?

src/background/services/grantBalance.ts
```diff
@@ -0,0 +1,183 @@
+import type {
+  WalletAddress,
+  OutgoingPaymentGrantSpentAmounts,
+} from '@interledger/open-payments';
+import type { Cradle } from '@/background/container';
+import {
+  GRANT_SPENT_AMOUNTS_SUPPORT_RECHECK_ALARM,
```
sidvishnoi: Nit: can keep this constant limited to this file, as not used anywhere else.

src/background/services/grantBalance.ts
```diff
@@ -0,0 +1,183 @@
...
+    this.logger = logger;
+    this.outgoingPaymentGrantService = outgoingPaymentGrantService;
+    this.browser = browser;
+
+    this.initialize();
+  }
+
+  private initialize() {
+    void this.registerBalanceUpdateHandler();
+  }
```
sidvishnoi: Let's do `initialize()` within `start()`, so the constructor stays pure. Helps with unit tests often.

src/shared/types.ts
```diff
@@ -105,12 +105,17 @@ export interface Storage {
...
+  supportsGrantSpentAmounts?: {
+    /**
+     * If the current wallet address' resource server supports the "spent amounts" endpoint
+     * for grants to not require polling of outgoing payments.
+     * @default undefined implies that support is unknown and needs to be checked.
+     */
+    supported?: boolean;
+
+    /** Timestamp of the last time we checked grant spent amounts support. */
+    lastCheckedAt?: number;
```
sidvishnoi: If the key is set, I think we always set both values together.

```suggestion
    supported: boolean;

    /** Timestamp of the last time we checked grant spent amounts support. */
    lastCheckedAt: number;
```

src/background/services/grantBalance.ts
```diff
@@ -0,0 +1,183 @@
...
+    } catch (error) {
+      if (isNotFoundError(error)) {
+        this.logger.debug(
+          'Resource server does not support grant spent amounts endpoint',
+          error,
+        );
+
+        await this.storage.set({
+          supportsGrantSpentAmounts: {
+            supported: true,
```
sidvishnoi:
```suggestion
            supported: false,
```

src/background/services/grantBalance.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    const updateBalance = async () => {
+      await this.saveUpdatedBalance();
+
+      const { continuousPaymentsEnabled } = await this.storage.get([
+        'continuousPaymentsEnabled',
+      ]);
+      const timeoutMs = continuousPaymentsEnabled
+        ? 1 * 60 * 1000
+        : 5 * 60 * 1000;
```
sidvishnoi: Should this be recomputed in `Timeout` callback?

src/background/services/grantBalance.ts
```diff
@@ -0,0 +1,183 @@
...
+    const spentAmounts = await this.getGrantSpentAmounts(walletAddress);
+    return spentAmounts?.spentDebitAmount !== undefined;
+  }
+
+  private async registerBalanceUpdateHandler() {
+    const { supportsGrantSpentAmounts } = await this.storage.get([
+      'supportsGrantSpentAmounts',
+    ]);
+
+    if (!supportsGrantSpentAmounts?.supported) {
```
sidvishnoi: So we don't accidentally register twice.
```suggestion
    if (!supportsGrantSpentAmounts?.supported || this.balanceUpdateTimeout) {
```

src/background/services/grantBalance.ts
```diff
@@ -0,0 +1,183 @@
...
+
+      if (!amounts?.spentDebitAmount) return;
+
+      this.storage.setSpentAmount(
+        this.outgoingPaymentGrantService.grantType,
+        amounts.spentDebitAmount.value,
+      );
+    } catch (error) {
+      this.logger.warn('Background balance update failed', error);
+      this.balanceUpdateTimeout?.clear();
```
sidvishnoi, line 183: Might want to set it to null as well? Or reset timeout?

src/shared/types.ts
MichaelKohler: That is indeed true for `lastCheckedAt` but for `supported` it can stay `undefined` if there is an error. We do pass it in those cases as well, would you prefer an explicit `boolean | undefined` instead?

sidvishnoi: Either works

sidvishnoi (approved):

PR #1425, opened 2026-07-14
fix: limit max custom rate exceptions to 200

sidvishnoi:
## Context

Finishing up #1345
Don't want too many exceptions - bad for user experience as well as extension storage.

- Limit to 200 (`MAX_CUSTOM_RATE_EXCEPTIONS`)
- Handle limit in manage exceptions screen
- Disable "add current site exception" input if reached limit
   - Ideally, we want a better error handling in this debounced input, but for now, this is ok.

### DIFF
--- a/src/background/config.ts
+++ b/src/background/config.ts
@@ -1,5 +1,4 @@
 export const DEFAULT_SCALE = 2;
-export const DEFAULT_INTERVAL_MS = 3_600_000;
 
 export const DEFAULT_BUDGET = '500';
 export const DEFAULT_RATE_OF_PAY = '60';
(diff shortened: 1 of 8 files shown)


src/background/services/background.ts
```diff
@@ -321,6 +322,10 @@ export class Background {
           if (rate === null) {
             await this.rateList.deleteRate(hostname);
           } else {
+            const count = await this.rateList.count();
+            if (count === MAX_CUSTOM_RATE_EXCEPTIONS) {
+              throw new Error('Cannot add further custom rates');
```
sidvishnoi, line 327: This error is silently dropped in `RateOfPay.tsx`, but handled by `sitesRateOfPay.length` check.

src/background/config.ts
```diff
@@ -1,5 +1,4 @@
 export const DEFAULT_SCALE = 2;
-export const DEFAULT_INTERVAL_MS = 3_600_000;
```
sidvishnoi, line 2: Unused.

MichaelKohler (approved):

PR #1430, opened 2026-07-15
fix(i18n): make i18n keys consistent

sidvishnoi:
## Context

Keys in locales file became inconsistent over time.

Use this grammar everywhere, no exceptions: `{feature}_{name}_{type}`
- `type` can be one of: title, desc, text, label, action, error, warn. Removed state/msg (folded into text).
- For numbered items: `{name}_{type}__{index}` (double underscore)

Next up: find hardcoded strings in app, and add them to locales file.

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -2,496 +2,501 @@
   "appDescription": {
     "message": "Support content you love. Pay as you browse!"
   },
-  "icon_state_monetizationActive": {
+  "icon_monetizationActive_text": {
     "message": "Active"
   },
-  "icon_state_monetizationInactive": {
+  "icon_monetizationInactive_text": {
     "message": "Inactive"
   },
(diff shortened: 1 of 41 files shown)


MichaelKohler (approved):

MichaelKohler:
> type can be one of: title, desc, text, label, action, error, warn

We are not enforcing this apart from `error` and `warn`, right?

sidvishnoi:
> > type can be one of: title, desc, text, label, action, error, warn
> 
> We are not enforcing this apart from `error` and `warn`, right?

Yes. We use `error` and `warn` with `ErrorWithKey`. For the rest, we don't have specific functions/components. e.g. `text` can be used with paragraphs, list items. Ideally, `action` will only be used with buttons, but we've more buttons, and often, buttons end up using `React.Node` instead of plain text.

PR #1432, opened 2026-07-20
feat(bg/grantBalance): optimistically update balance on payments

### LINKED ISSUE #1431: Update balance optimistically after payments
With https://github.com/interledger/web-monetization-extension/pull/1423 we introduced a regular balance update when opening the popup, as well as while the popup is open. However this can still feel odd, as when doing a one-time payment, the balance will not necessarily be up to date right away, such as when `grantSpentDebitAmount` is not supported by the wallet provider or when there are several requests in-flight.

We want to always optimistically update the balance whenever a payment is initiated. If the payment fails, we adjust the amount back to the correct amount.

MichaelKohler:
## Context

Closes #1431.

This introduces optimistic balance updates. We deduct the amount when initiating the action, and then correct it as necessary if there are failures.

### DIFF
--- a/src/background/services/__tests__/grantBalance.test.ts
+++ b/src/background/services/__tests__/grantBalance.test.ts
@@ -1,0 +1,508 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { OpenPaymentsClientError } from '@interledger/open-payments';
+import { BACKGROUND_TO_POPUP_CONNECTION_NAME } from '@/shared/messages';
+import type { Storage } from '@/shared/types';
+import { EventsService } from '../events';
+import {
+  GrantBalanceService,
+  GRANT_SPENT_AMOUNTS_SUPPORT_RECHECK_ALARM,
+  GRANT_SPENT_AMOUNTS_SUPPORT_RECHECK_INTERVAL_MS,
+} from '../grantBalance';
+import { makeBrowser, makeLogger, makeStorage, walletInfo } from './helpers';
+
+beforeEach(() => {
+  vi.useFakeTimers();
+});
+
+afterEach(() => {
+  vi.useRealTimers();
+});
+
+function makeService(
+  overrides: {
+    isAnyGrantUsable?: boolean;
+    grantType?: 'recurring' | 'one-time';
+    storageState?: Partial<Storage>;
+  } = {},
+) {
+  const storage = makeStorage(overrides.storageState);
+  const browser = makeBrowser();
+  const events = new EventsService();
+  const logger = makeLogger();
+  const outgoingPaymentGrantService = {
+    grantType: overrides.grantType ?? 'recurring',
+    isAnyGrantUsable: overrides.isAnyGrantUsable ?? false,
+    getGrantSpentAmounts: vi.fn(),
+  };
+  const service = new GrantBalanceService({
+    storage,
+    logger,
+    events,
+    outgoingPaymentGrantService,
+    browser,
+  } as unknown as ConstructorParameters<typeof GrantBalanceService>[0]);
+  return {
+    service,
+    storage,
+    browser,
+    events,
+    outgoingPaymentGrantService,
+  };
+}
+
+function makePort(name: string) {
+  const disconnectListeners: Array<() => void> = [];
+  return {
+    name,
+    error: undefined,
+    onDisconnect: {
+      addListener: (cb: () => void) => disconnectListeners.push(cb),
+      removeListener: vi.fn(),
+    },
+    disconnect: () => {
+      for (const cb of disconnectListeners) cb();
+    },
+  };
+}
+
+describe('open_payments.outgoing_payment_created event handling', () => {
+  async function emitAdjustment(events: EventsService, amount: string) {
+    events.emit('open_payments.outgoing_payment_created', {
+      debitAmount: { value: amount, assetCode: 'USD', assetScale: 2 },
+    });
+    await vi.advanceTimersByTimeAsync(1000);
+  }
+
+  it('adds amount to the recurring grant spent amount', async () => {
+    const { service, storage, events } = makeService({
+      grantType: 'recurring',
+      storageState: { recurringGrantSpentAmount: '100' },
+    });
+    service.start();
+
+    await emitAdjustment(events, '25');
+
+    expect(storage.setSpentAmount).toHaveBeenCalledWith('recurring', '125');
+  });
+
+  it('adds amount to the one-time grant spent amount', async () => {
+    const { service, storage, events } = makeService({
+      grantType: 'one-time',
+      storageState: { oneTimeGrantSpentAmount: '50' },
+    });
+    service.start();
+
+    await emitAdjustment(events, '10');
+
+    expect(storage.setSpentAmount).toHaveBeenCalledWith('one-time', '60');
+  });
+
+  it('accumulates correctly across multiple sequential emits', async () => {
+    const { service, storage, events } = makeService({
+      grantType: 'recurring',
+      storageState: { recurringGrantSpentAmount: '0' },
+    });
+    service.start();
+
+    await emitAdjustment(events, '10');
+    await emitAdjustment(events, '5');
+    await emitAdjustment(events, '-3');
+
+    expect(storage.setSpentAmount).toHaveBeenLastCalledWith('recurring', '12');
+    expect(storage.state.recurringGrantSpentAmount).toBe('12');
+  });
+});
+
+describe('checkGrantSpentAmountsSupport (via start())', () => {
+  it('does nothing when no grant is usable', async () => {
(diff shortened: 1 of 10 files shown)


src/background/services/paymentManager.ts
```diff
@@ -215,6 +217,9 @@ export class PaymentManager {
         amount: amount.toString(),
       })),
     });
+    this.deps.events.emit('balance.adjust_spent_amount', {
+      amount: payingAmount.toString(),
```
MichaelKohler: It does feel kinda odd to say I want to pay `MXN 0.50` but then I get deducted `MXN 0.36`. But deducting 50 first and then adding it back is not better and has more visual movement. Have we considered in the past to introduce an explanation why the requested amount can't be paid as-is?

sidvishnoi (commented):
Wondering, if we instead emit a payment created event which the balance service listens to, and adapts to optimistic balance update (summing debitAmount). It might be good to debounce/batch it a few ms.
It certain payments fail, we can update the balance in next run so user doesn't feel like balance dancing in a single payment - using "payment completed" events.
The idea is, right now, emitting this balance adjust event is adding logic to services that don't care about it. Emitting payment create/complete from payment session (or op) service is a blind event. That event can be used for a per-site stats feature as well in future.

src/background/services/paymentManager.ts
```diff
@@ -293,6 +300,12 @@ export class PaymentManager {
       0n,
     );
 
+    if (debitAmount < expectedDebitAmount) {
+      this.deps.events.emit('balance.adjust_spent_amount', {
+        amount: (expectedDebitAmount - debitAmount).toString(),
```
kjmitchelljr: If I'm following the flow correctly, when it comes to the polling that takes place in `getPayStatus` should this be flipped to: 
`amount: (debitAmount - expectedDebitAmount).toString()`?

MichaelKohler: In this case here we debited less than we previously deducted. So we need to _add_ the difference back to the balance.

src/background/services/paymentSession.ts
```diff
@@ -503,16 +501,27 @@ export class PaymentSession {
...
+          this.events.emit('open_payments.outgoing_payment_completed', {
+            debitAmount: outgoingPayment.debitAmount,
+            sentAmount: outgoingPayment.sentAmount,
+            status: 'failed',
+          });
+
+          if (outgoingPayment.sentAmount.value === '0') {
+            throw new ErrorWithKey('pay_outgoingPaymentFailed_error');
+          }
         }
```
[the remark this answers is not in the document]
MichaelKohler: Yes, but first need to figure out how we want to handle partial success/failure.

src/background/services/grantBalance.ts
```diff
@@ -44,6 +54,48 @@ export class GrantBalanceService {
     });
 
     void this.checkGrantSpentAmountsSupport();
+
+    this.events.on(
+      'open_payments.outgoing_payment_created',
+      async ({ debitAmount }) => {
```
sidvishnoi: Nit: 
```suggestion
      ({ debitAmount }) => {
```

src/background/services/grantBalance.ts
```diff
@@ -44,6 +54,48 @@ export class GrantBalanceService {
...
+        const amount = BigInt(debitAmount.value);
+        this.adjustSpentAmount.enqueue(amount);
+      },
+    );
+
+    this.events.on(
+      'open_payments.outgoing_payment_completed',
+      async ({ debitAmount, sentAmount, status }) => {
```
sidvishnoi: Nit: 
```suggestion
      ({ debitAmount, sentAmount, status }) => {
```

sidvishnoi (commented):
Me likey. Will test locally, overall very happy with the architecture. Copilot comments sensible as well.

sidvishnoi:
Working very well in general :rocket: 

Found a scenario where balance doesn't update correctly without `getGrantSpentAmounts`:

https://github.com/user-attachments/assets/38c2c4f3-e359-439a-a925-b53c3d31fe37


Connected wallet: https://ilp.interledger-test.dev/sid (USD)
Receivers: https://webmonetization.org/play/?wa=https://ilp.interledger-test.dev/sid-zar-1&wa=https://ilp.interledger-test.dev/sid-mxn-1

Continuous payments disabled. Sent $0.50. Balance didn't update fully during polling step, but was corrected when popup re-opened. Following are some messages/logs from background:
```
[main] Received message PAY_WEBSITE  Object { amount: "0.50" }					background.ts:231:17
[payment-manager] webmonetization.org pay(50) to 2 sessions							paymentManager.ts:194:17
[payment-manager] webmonetization.org sending outgoing payments					paymentManager.ts:206:17
	Object { amount: {…}, distribution: (2) […] }
	 amount: Object { total: "50", paying: "50", remaining: "0" }
	 distribution: Array [ {…}, {…} ]
	  0: Object { id: "2c65466b-de5d-4a97-97ff-52349c7a42ac", receiver: "https://ilp.interledger-test.dev/sid-zar-1", amount: "25" }
	  1: Object { id: "d87394ac-4d00-4dc6-865d-e920300930bf", receiver: "https://ilp.interledger-test.dev/sid-mxn-1", amount: "25" }
	  length: 2
```

Sometimes it computes correctly, but sometimes doesn't.
I tried sending $0.45, it reduced balance by $0.48 (after polling complete msg), then corrected it (reduce by $0.45) later popup-open.

src/background/services/__tests__/rateList.test.ts
```diff
@@ -86,14 +86,13 @@ afterEach(() => {
...
+  wallet: WalletInfo | null = makeWallet(),
 ): RateListService {
-  const storage = {
-    get: vi.fn().mockResolvedValue({ walletAddress: wallet }),
-  } as unknown as StorageService;
-  return new RateListService({ storage });
+  const storage = makeStorage({ walletAddress: wallet });
+  return new RateListService({
+    storage,
+  } as unknown as ConstructorParameters<typeof RateListService>[0]);
```
MichaelKohler, line 95: I don't like this cast, in a quiet minute with a fresh mind I will look into that again (follow-up) to improve it.

src/background/services/__tests__/grantBalance.test.ts
```diff
@@ -0,0 +1,667 @@
...
+  async function emitOutgoingPaymentCreated(
+    events: EventsService,
+    amount: string,
+    grantType: GrantDetails['type'],
+  ) {
+    events.emit('open_payments.outgoing_payment_created', {
+      debitAmount: { value: amount, assetCode: 'USD', assetScale: 2 },
+      grantType,
+    });
+    await vi.advanceTimersByTimeAsync(1000);
```
sidvishnoi: Nit: I would recommend keeping this out of emit, and call it manually when needed. Makes flow easier to follow.

src/background/services/__tests__/grantBalance.test.ts
```diff
@@ -0,0 +1,667 @@
...
+      debitAmount: { value: '25', assetCode: 'USD', assetScale: 2 },
+      grantType,
+    });
+    await vi.advanceTimersByTimeAsync(1000);
+
+    expect(storage.setSpentAmount).toHaveBeenLastCalledWith(grantType, '50');
+    expect(
+      (await storage.get(['recurringGrantSpentAmount']))
+        .recurringGrantSpentAmount,
+    ).toBe('50');
```
sidvishnoi:
```suggestion
    await expect(storage.get(['recurringGrantSpentAmount'])).resolves.toEqual({ recurringGrantSpentAmount: '50' });
```

src/background/services/__tests__/grantBalance.test.ts
```diff
@@ -0,0 +1,667 @@
...
+      sentAmountValue: '10',
+      status: 'failed',
+      grantType,
+    });
+
+    expect(storage.setSpentAmount).toHaveBeenCalledWith(grantType, '85');
+  });
+});
+
+describe('checkGrantSpentAmountsSupport (via start())', () => {
```
sidvishnoi: Sending these tests (separate from current optimistic update changes) in a new PR would be preferable.

src/background/services/grantBalance.ts
```diff
@@ -44,6 +62,59 @@ export class GrantBalanceService {
...
+        } else if (grantType === 'one-time') {
+          this.adjustOneTimeSpentAmount.enqueue(amount);
+        }
+      },
+    );
+
+    this.events.on(
+      'open_payments.outgoing_payment_completed',
+      ({ debitAmount, sentAmount, status, grantType }) => {
+        if (status !== 'failed') {
```
sidvishnoi, line 84: Perhaps add a comment, why we can safely ignore success here (for future us).

sidvishnoi (approved):

sidvishnoi (approved):
Approve++ :shipit:

Issue #979, opened 2025-03-13
[FEATURE REQUEST]


ferdeline:
### Describe the feature you would like to request

I would like to request Internationalization (i18n) support to ensure the browser extension is accessible and relevant to a global audience. Currently, the extension’s user interface and onboarding experience are only available in English, which limits its adoption among non-English speakers, including those in emerging markets who stand to benefit most from open and inclusive financial tools.

### Describe the solution you would like

For a seamless user experience, the extension should automatically detect the user’s preferred language based on their browser settings. However, it should also provide an option in the settings for manual language selection, allowing users to switch to their preferred language at any time. This ensures flexibility, particularly in multilingual regions. To maximize impact, I propose that the first phase of localization should focus on key languages spoken in regions where Web Monetization adoption is expected to grow. Spanish, Portuguese, French, and Arabic would be excellent starting points, as they cover large populations in Latin America, Africa, and the Middle East. As Arabic does not use Latin characters and is a right-to-left language, this would help us ensure the browser extension can accommodate bidirectional text rendering. Over time, additional languages could be added, possibly with contributions from the Interledger community.

sidvishnoi:
This is something we intend to do in a future release. But no timeline for now.

We've architecture in place to support different languages (we have most of the strings in `_locales`), which means once we've translations in other languages, it'll switch automatically, depending on user's browser/OS settings.

On adding a language switcher: we'll see how other extensions do it. I do not have an example handy where extensions instead of using browser's i18n API, implement their own switchers, but technically should be doable. Browsers don't have a language switcher option for extensions in their UI either.

PR #1436, opened 2026-07-22
feat(i18n): basic markup in translation strings, add more strings

### LINKED ISSUE #1267: Explore markup-rich translations
Presently, we can only add plain text in localization strings. If we want to bold/emphasise some text or mark some text as `<code>`, we've few options: either use HTML or use a markdown processor, or manually split strings into HTML and plain text bits. Markdown appears best to go ahead with, but we need to convert markdown to React JSX, not HTML strings (so we can add onClick behavior to links/buttons, or other custom behavior, and also avoids HTML injection risks).

## Plan:

In locales:
```diff lang=json
- "message": "Uses PostHog to collect usage and analytics data."
+ "message": "Uses [PostHog](https://posthog.com/) to collect usage and analytics data."
```

In components that use `t = useTranslation()`:
```diff lang=jsx
- <p>{t('key')}</p>
+ <p>{md(t('key'), {
+   link: (text, url) => (<a href={url} className="...">{text}</a>) // when we want custom behavior
+ })}</p>
```

Such utility may also be helpful with our docs websites (that use JSX/Astro).

Further context: https://github.com/interledger/web-monetization-extension/pull/1263#discussion_r2588927718, https://github.com/interledger/web-monetization-extension/pull/1251#pullrequestreview-3530583330

sidvishnoi:
## Context

Closes https://github.com/interledger/web-monetization-extension/issues/1267
Paves way for https://github.com/interledger/web-monetization-extension/issues/979

- Support basic inline markup in i18n strings, powered by [`imd`](https://github.com/sidvishnoi/imd) package.
- Move many more hard-coded strings from code to `messages.json`

### Example:

Now, the entire string is in one place, and its rendering is handled separately. When we support more languages, placement of parameters might change, but we don't have to update our markup for those.

A good example can be `settings_budget_renewDate_text` - it has a `<time>` element In other language (Hindi for example), we can have something like:

```
 Your budget will renew on <time>$RENEW_DATE$</time>.    // en
 आपका बजट <time>$RENEW_DATE$</time> को रिन्यू होगा।        // hi
 aapka budget <time>$RENEW_DATE$</time> ko renew hoga.   // hi-Latn
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "date-fns": "^4.4.0",
     "http-message-signatures": "^1.0.6",
     "idb": "^8.0.3",
+    "imd": "npm:@sidvishnoi/imd@^0.0.2",
     "iso8601-duration": "^2.1.4",
     "loglevel": "^1.9.2",
     "posthog-js": "1.396.1",
@@ -59,6 +60,7 @@
     "@types/react": "^19.2.17",
     "@types/react-dom": "^19.2.3",
     "@types/webextension-polyfill": "^0.12.5",
+    "@vitest/coverage-v8": "^4.1.10",
     "archiver": "^8.0.0",
     "esbuild": "^0.28.1",
     "esbuild-node-builtin": "^0.1.1",
@@ -73,8 +75,7 @@
     "totp-generator": "2.0.1",
     "typescript": "npm:typescript@^6.0.3",
     "typescript7": "npm:typescript@^7.0.2",
-    "vitest": "^4.1.10",
-    "@vitest/coverage-v8": "^4.1.10"
+    "vitest": "^4.1.10"
   },
   "engines": {
     "pnpm": "^11.9.0",
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -482,12 +482,6 @@
   "postConnect_title": {
     "message": "Web Monetization extension"
   },
-  "postConnect_empty_text": {
-    "message": "Nothing to see here"
-  },
-  "postConnect_unmapped_text": {
-    "message": "No mapping from status to messages!!"
-  },
   "postConnect_retry_action": {
     "message": "Try again"
   },
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -2,6 +2,7 @@
   "appDescription": {
     "message": "Support content you love. Pay as you browse!"
   },
+  "app_loading_text": { "message": "Loading" },
   "icon_monetizationActive_text": {
     "message": "Active"
   },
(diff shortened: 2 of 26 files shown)


package.json
```diff
@@ -31,6 +31,7 @@
     "date-fns": "^4.4.0",
     "http-message-signatures": "^1.0.6",
     "idb": "^8.0.3",
+    "imd": "npm:@sidvishnoi/imd@^0.0.2",
```
MichaelKohler, line 34: What's the reason for not using `@sidvishnoi/imd` directly?

src/_locales/en/messages.json
```diff
@@ -418,10 +467,7 @@
     "message": "Extension Permissions"
   },
   "postInstallConsent_permissions_text": {
-    "message": "The extension needs basic permissions to work."
-  },
-  "postInstallConsent_permissions_link_text": {
-    "message": "See details."
+    "message": "The extension needs basic permissions to work. [See details](https://github.com/interledger/web-monetization-extension/blob/main/docs/PERMISSIONS.md)."
```
MichaelKohler, line 470: We IMHO shouldn't link to GitHub docs, but rather document it on webmonetization.org. Maybe we could add a page but not expose it in the sidebar? (not to be done with this PR though)

src/_locales/en/messages.json
```diff
@@ -302,6 +342,9 @@
   "postInstall_title": {
     "message": "Let's get you started with Web Monetization!"
   },
+  "postInstall_step_text": {
+    "message": "Step "
```
MichaelKohler, line 346: Have you encountered other similar cases which we currently do not support yet? Might be interesting to explore solutions for this one, but definitely not worth it right now.

src/_locales/en/messages.json
```diff
@@ -433,6 +479,21 @@
...
   },
 
+  "postConnect_title": {
+    "message": "Web Monetization extension"
+  },
+  "postConnect_empty_text": {
+    "message": "Nothing to see here"
+  },
+  "postConnect_unmapped_text": {
+    "message": "No mapping from status to messages!!"
```
MichaelKohler: Also not for this PR: this assumes that we're gonna notice this. Probably would make sense to keep this more generic as an "unknown" state message.

MichaelKohler (approved):

package.json
sidvishnoi: Too much sidvishnoi 😂

src/_locales/en/messages.json
sidvishnoi: Ideally yes. But one benefit of GitHub based ones is, they can tied to release, while site will only be about latest.

src/_locales/en/messages.json
sidvishnoi: I think this was the only one. One option is to add two strings - one for sr-only and other one.

src/_locales/en/messages.json
sidvishnoi: It was more of an internal error state. I think I'll remove these from messages and just log and redirect to home page.

Issue #1081, opened 2025-06-10
[Testing] Add local test server for E2E tests


sidvishnoi:
Presently, we rely on [https://webmonetization.org/play/](<https://webmonetization.org/play/>) to add link tags on a page for automated testing. This has some issues:

* Playground includes lots of other things which add up as noise in testing (clicking button to add link etc. although the `?wa` URL params were helpful)
* Playground can only dynamically add link tags with JS, in a single manner.
* We can't have custom CSP or other headers there
* Even if small, there's a performance overhead on using a remote playground website.

We want to set up a local server of test pages that we can use in the extension.<br>The test pages will be version controlled along with the extension, and also serve as test pages for other implementations and users.

In future, we might want to run the test server on additional domains (for CSP/Permissions Policy specific testing, instead of just `localhost:{port}`.

Aside: [https://github.com/interledger/web-monetization-extension/pull/1027](<https://github.com/interledger/web-monetization-extension/pull/1027>) enabled us to run the extension on localhost.

MichaelKohler:
I'm gonna close this as with the new focus this does not seem like something that would be super beneficial of having contributor time spent on it.

PR #1440, opened 2026-07-27
test(e2e): improve E2E test flakiness

MichaelKohler:
## Context
Preparations before starting to work on #1081, so that we do not have failing tests deterring us.

It seems that the clock is not faked in the background script. Additionally tests depend on network conditions, and sometimes time out due to that. Therefore I'm adjusting timeouts to be less likely for tests to be flaky.

Additionally the rate of pay for the overpaying test is adjusted to not give race conditions with `waitForTimeout`, with payments sneaking into the window we test for.


tests/e2e/multiple.spec.ts
```diff
@@ -138,10 +133,12 @@ test.describe('should monetized site with multiple wallet address', () => {
...
   test('different currencies', async ({ page, popup, background, context }) => {
-    // At rateOfPay=3600, we pay $0.01 every second. But given minimum interval
-    // is 2s, we'll need to wait for 2s.
+    // PaymentManager's round-robin scheduler only pays a session once enough
+    // pendingAmount has accumulated to clear that *session's* probed `minSendAmount`. A
+    // session may need several ticks before it gets paid.
+    test.setTimeout(120_000);
     const rateOfPay = '3600';
-    const interval = MIN_PAYMENT_WAIT;
+    const crossCurrencyTimeout = 30_000;
```
sidvishnoi, line 140: We're setting a high rate of pay, so the number of ticks is reduced. The minimum interval then becomes more relevant than ticks.

tests/e2e/multiple.spec.ts
```diff
@@ -138,10 +133,12 @@ test.describe('should monetized site with multiple wallet address', () => {
   });
 
   test('different currencies', async ({ page, popup, background, context }) => {
-    // At rateOfPay=3600, we pay $0.01 every second. But given minimum interval
-    // is 2s, we'll need to wait for 2s.
+    // PaymentManager's round-robin scheduler only pays a session once enough
+    // pendingAmount has accumulated to clear that *session's* probed `minSendAmount`. A
+    // session may need several ticks before it gets paid.
+    test.setTimeout(120_000);
```
sidvishnoi: Nit: Will a `test.slow()` be enough? Or this time is a better estimate?

tests/e2e/overpaying.spec.ts
```diff
@@ -14,10 +15,15 @@ test.afterEach(({ context }) => {
...
 test.describe('should not pay immediately when overpaying', () => {
-  test('on page reload', async ({ page, popup, context }) => {
+  test('on page reload', async ({ page, popup, background, context }) => {
     const { outgoingPaymentCreatedCallback } =
       interceptPaymentCreateRequests(context);
 
+    await background.evaluate(
+      (rateOfPay) => chrome.storage.local.set({ rateOfPay }),
+      DEFAULT_RATE_OF_PAY,
+    );
```
sidvishnoi, line 25: I added a util in siteRate.spec.ts for this. Let's use that.

tests/e2e/multiple.spec.ts
```diff
@@ -79,7 +77,6 @@ test.describe('should monetized site with multiple wallet address', () => {
         paymentPointer: walletAddresses[0],
       });
 
-      await context.clock.runFor(interval);
```
sidvishnoi, line 82: TODO(sidvishnoi): I'll investigate this in detail. I know the clock doesn't work with background right now, but I kept these simulations hoping we can fix it some time later. It could be something with how our Timeout class is defined, or maybe how playwright deals with extensions.

tests/e2e/overpaying.spec.ts
```diff
@@ -14,10 +15,15 @@ test.afterEach(({ context }) => {
...
 
 test.describe('should not pay immediately when overpaying', () => {
-  test('on page reload', async ({ page, popup, context }) => {
+  test('on page reload', async ({ page, popup, background, context }) => {
     const { outgoingPaymentCreatedCallback } =
       interceptPaymentCreateRequests(context);
 
+    await background.evaluate(
+      (rateOfPay) => chrome.storage.local.set({ rateOfPay }),
+      DEFAULT_RATE_OF_PAY,
```
sidvishnoi, line 24: Why do we need to set the default though? Is the default value different here? It's set at wallet connection time (which happens on each test).

tests/e2e/multiple.spec.ts
sidvishnoi: Likely this https://github.com/microsoft/playwright/issues/31772

tests/e2e/overpaying.spec.ts
MichaelKohler: Which util do you mean? I think setting the global rate here is simpler than creating a custom hostname based rate here.

tests/e2e/overpaying.spec.ts
MichaelKohler: That rate of pay is based on the budget recommendation JSON. Here we want to make sure that we do not have payments during the test that might change the callback counts. `DEFAULT_RATE_OF_PAY` results in an interval that is slow enough to not influence the test.

tests/e2e/multiple.spec.ts
MichaelKohler: Good point, didn't know that existed. It is enough.

tests/e2e/overpaying.spec.ts
sidvishnoi: https://github.com/interledger/web-monetization-extension/blob/c3363c477aacaa610a9a44f0d501dc0787a064f9/tests/e2e/siteRate.spec.ts#L302-L309

sidvishnoi (approved):

tests/e2e/helpers/common.ts
```diff
@@ -166,3 +167,12 @@ export function interceptPaymentCreateRequests(context: BrowserContext) {
     incomingPaymentCreatedCallback,
   };
 }
+
+export async function defineGlobalRate(
+  background: Worker,
```
sidvishnoi: Nit: use `Background` instead of `Worker`.

sidvishnoi (approved):

PR #1442, opened 2026-07-27
refactor(background): remove deduplicator as a service, keep fn

sidvishnoi:
## Context

Deduplicator doesn't belong as a service. It's just a regular function, like throttle/debounce.

- Remove Deduplicator as service, add a `dedupe` function to utils
- Remove argument caching support, as unused (since 2y)
- Also removed logging when dedupe is called


docs/DEVELOP.md
```diff
@@ -199,8 +199,6 @@ mindmap
   - The backend part is for orchestration only.
 - **`EventsService`**:
   - A simple EventEmitter/pub-sub service
-- **`Deduplicator`**:
```
sidvishnoi, line 202: @MichaelKohler Should update to include `grantBalance` service..

MichaelKohler: Will do!

MichaelKohler (approved):

PR #1468, opened 2026-08-20
docs(testingplan): adjust phrasing, update wallet providers and add new test case

MichaelKohler:
Our testing plan needed some updates as per latest announcement, and also some general fixes.

* Removed dedicated wallet provider list as we already link to the list in the docs
* Adjusted environment/requirements to be up to date
* Several phrasing adjustments
* Removed Fynbos mentions
* Added custom rate test case

### DIFF
--- a/cspell-dictionary.txt
+++ b/cspell-dictionary.txt
@@ -82,6 +82,7 @@ storytogo
 ericahargreave
 roamancing
 naturallyours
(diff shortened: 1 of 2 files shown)


sidvishnoi (approved):
I don't know if we want to maintain this doc, but changes look good.

PR #1469, opened 2026-08-21
fix(polyfill): expose `MonetizationCurrencyAmount` without constructor

### LINKED ISSUE #501: Expose `MonetizationCurrencyAmount`
Spec says it's exposed on Window.
Also, make sure `amountSent` is an instance of `MonetizationCurrencyAmount`.

MichaelKohler:
## Context

Closes #501.

This exposes `MonetizationCurrencyAmount` to `window`. We could use [LegacyNoInterfaceObject](https://webidl.spec.whatwg.org/#LegacyNoInterfaceObject), however it [should not be used for new APIs](https://webidl.spec.whatwg.org/#js-legacy-extended-attributes). Therefore we expose it, but do not allow construction.

### DIFF
--- a/scripts/esbuild/prod.ts
+++ b/scripts/esbuild/prod.ts
@@ -89,35 +89,45 @@ function zipPlugin({
 }
 
 /**
- * Unmangles the MonetizationEvent class
+ * Unmangles the MonetizationEvent and MonetizationCurrencyAmount classes
  */
 function preservePolyfillClassNamesPlugin({
   outDir,
 }: {
   outDir: string;
 }): ESBuildPlugin {
+  const classNames = ['MonetizationEvent', 'MonetizationCurrencyAmount'];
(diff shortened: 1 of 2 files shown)


sidvishnoi (approved):
You might want to make sure the class has right name after bundling, otherwise it'll look like
```
amountSent: g { currency: 'USD', value: '1.2' }
```
instead of 
```
amountSent: MonetizationCurrencyAmount { currency: 'USD', value: '1.2' }
```
in devtools etc.

https://github.com/interledger/web-monetization-extension/blob/47379a697557d4820bf7d9db76b1b8cb7f2a65f3/scripts/esbuild/prod.ts#L94

Existing `definitionRegex` is safe with this PR (due to `extends Event` suffix).

sidvishnoi (approved):

PR #1489, opened 2026-09-17
feat(polyfill): add granular handling in polyfill

### LINKED ISSUE #1482: [Native WM] Step 1 - Adjust polyfill to be more granular
Our plan is to [move existing usage](https://docs.google.com/document/d/1-B2DBzv7eqsX7jugZAXRKB1xeAIy37phqwuObxO_d6s/edit?tab=t.0) from the extension over to the native functionality when available in the browser. For this, we need to adjust the polyfill, as different parts will land at different times in browsers.

- Remove initial early return
- Only patch supports if it doesn’t already return true
- Do not patch `MonetizationEvent` if it already exists
- Do not patch `MonetizationCurrencyAmount` if it already exists
- Do not patch `onmonetization` if it already exists
- Do not add internal monetization event listeners if it already exists (avoid re-registration though)
  - We can use a Symbol on window to indicate that it was already registered. This seems cleaner to me than saving the event handler function on window and relying on the browser internal de-dupe.

MichaelKohler:
## Context

Closes #1482.

This makes the polyfill more granular, checking if each piece already exists before overwriting it. This is required for the in-browser implementation rollout, as otherwise the extension would keep overwriting the browser-internal implementation.


MichaelKohler:
@sidvishnoi quick ping for the review :)

sidvishnoi (approved):
LGTM. Worth testing in respect to https://github.com/interledger/web-monetization-extension/issues/418 as well - if it improves or worsens the situation.

PR #1490, opened 2026-09-18
fix(polyfill): add walletAddress to MonetizationEvent

MichaelKohler:
## Context

Follow-up to https://github.com/WICG/webmonetization/pull/520.

This adds `walletAddress` to the `MonetizationEvent` dispatch. For now we keep the `paymentPointer` though.


src/content/polyfill.ts
```diff
@@ -93,19 +93,22 @@ import type { MonetizationEventPayload } from '@/shared/messages';
...
       super(type, { bubbles: true });
-      const { amountSent, incomingPayment, paymentPointer } = eventInitDict;
+      const { amountSent, incomingPayment, paymentPointer, walletAddress } =
+        eventInitDict;
       this.amountSent = createMonetizationCurrencyAmount(
         amountSent.currency,
         amountSent.value,
       );
       this.incomingPayment = incomingPayment;
       this.paymentPointer = paymentPointer;
```
sidvishnoi: Let's make this a getter and emit deprecation warning (once) on access.

src/shared/messages.ts
```diff
@@ -312,6 +312,7 @@ export interface MonetizationEventDetails {
   amountSent: PaymentCurrencyAmount;
   incomingPayment: OutgoingPayment['receiver'];
   paymentPointer: WalletAddress['id'];
```
sidvishnoi: Can remove this from message/backend entirely. Just need to define in MonetizationEvent, otherwise it'll create confusion.

tests/e2e/multiple.spec.ts
```diff
@@ -74,30 +74,34 @@ test.describe('should monetized site with multiple wallet address', () => {
       await expect(monetizationCallback).toHaveBeenCalledTimes(1);
       await expect(monetizationCallback).toHaveBeenLastCalledWithMatching({
         paymentPointer: walletAddresses[0],
+        walletAddress: walletAddresses[0],
```
sidvishnoi: It's ok to keep `paymentPointer` only in `simple.spec` (to ensure it's there), we can do a replacement for the rest. We don't need `paymentPointer` in every event test. Keep it simple/clean.

src/content/polyfill.ts
```diff
@@ -120,8 +136,9 @@ import type { MonetizationEventPayload } from '@/shared/messages';
         console.warn(msg);
         eventDetailDeprecationEmitted = true;
       }
-      const { amountSent, incomingPayment, paymentPointer } = this;
-      return { amountSent, incomingPayment, paymentPointer };
+      const { amountSent, incomingPayment, paymentPointer, walletAddress } =
+        this;
+      return { amountSent, incomingPayment, paymentPointer, walletAddress };
```
sidvishnoi, line 141: `detail` doesn't need `walletAddress` (it was already legacy).

sidvishnoi (approved):

src/content/polyfill.ts
MichaelKohler: I found it confusing if `details` is not aligned, even if it's legacy.

